PR #6, opened 2015-03-11
replace getColumnName() with getColumnLabel() 

torihat:
> replace getColumnName() with getColumnLabel() , as getColumnName() of mysql-connector returns original column name.

    query: select name as name2 from t

mysql
- getColumnName() -> "name"
- getColumnLabel() -> "name2"

postgresql
- getColumnName() -> "name2"
- getColumnLabel() -> "name2"


frsyuki:
:+1:

PR #7, opened 2015-04-11
use getColumnLabel to know column name

takumakanari:
When we config raw query  *select x as x_name, y as y_name*  in config file, column name will be 'x' and 'y' instead of 'x_name', 'y_name'.
This PR will resolve this problem by using method *getColumnLabel*.

Can you review & merge this PR if this behavior is correct?


takumakanari:
#6 looks same modification, I'll close this PR.

PR #11, opened 2015-07-03
Support SQL Server

hito4t:


frsyuki:
Great job! Thank you.
Apparently, we can't change schema search path at SQL Server per connection. Could you check pull-req and merge?: https://github.com/embulk/embulk-input-jdbc/pull/12

frsyuki:
For full schema support, I created #13.

PR #15, opened 2015-07-07
Support embulk 0.6.16

hito4t:
- PluginClassLoader class was moved. (0.6.16)
- Timestamp format in column_options are no longer used in csv formatter. (0.6.14)


frsyuki:
:+1:

Issue #16, opened 2015-08-03
embulk-input-oracle stalls on SocketInputStream.read


frsyuki:
Apparently, timeout (SO_RCVTIMEO) is not set
Timeout can be set by default safely because most of operations are designed to be idempotent and have retrying.

Stacktrace:
https://gist.github.com/tasuku-s/6d98b2091879b686ec91

hito4t:
`MySQLInputPlugin`, `PostgreSQLInputPlugin` and `RedshiftInputPlugin` set connection(login) time out and socket timeout, but `OracleInputPlugin` doesn't set. `OracleInputPlugin` should do.

`java.sql.DriverManager` has `setLoginTimeout` method, but it may be inappropriate because it is a static method.

`java.sql.Statement` has `setQueryTimeout` method. By the javadoc,
<pre>
A JDBC driver must apply this limit to the execute, executeQuery and executeUpdate methods.
Note: JDBC driver implementations may also apply this limit to ResultSet methods (consult your driver vendor documentation for details). 
</pre>

I've checked the javadoc of Oracle JDBC Driver, but nothing is written about `setQueryTimeout`.

By the javadoc of Oracle JDBC Driver, it has following connection properties.
They may work instead of standard JDBC Driver methods (`setLoginTimeout` and `setQueryTimeout`). 
http://docs.oracle.com/cd/E11882_01/appdev.112/e13995/oracle/jdbc/OracleConnection.html#CONNECTION_PROPERTY_THIN_NET_CONNECT_TIMEOUT

http://docs.oracle.com/cd/E11882_01/appdev.112/e13995/oracle/jdbc/OracleConnection.html#CONNECTION_PROPERTY_THIN_READ_TIMEOUT

Now connection(login) timeout limit (300s) and socket timeout limit (1800s) are written directly on XXXPlugin class, but they should be gotten by config.

And SQLServerInputPlugin should also set timeouts.

hito4t:
SQL Server has connection timeout property but doesn't have socket timeout property.
https://msdn.microsoft.com/ja-jp/library/ms378988%28v=SQL.110%29.aspx

embulk-input-sqlserver should set `Statement#setQueryTimeout` instead of socket timeout.

PR #17, opened 2015-08-14
Add connect_timeout and socket_timeout to config

hito4t:
- Add connect_timeout and socket_timeout to config.
- MySQLInputPlugin, PostgreSQLInputPlugin and RedshiftInputPlugin had fixed timeout values. Now they get timeout values from config.
- embulk-input-oracle and embulk-input-sqlserver didn't support timeout (see #16 ). Now they support.
- embulk-input-jdbc only support socket_timeout (actually query timeout) using standard JDBC driver method (`Statement#setQueryTimeout`).It doesn't support connect_timeout because the standard JDBC driver method for it (`DriverManager#setLoginTimeout`) is static and it may affect other modules.


frsyuki:
Nice :+1:

PR #24, opened 2015-10-13
Use jacoco plugin

muga:
Let's use JaCoCo plugin to check test coverage.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,18 +4,23 @@ plugins {
     id 'com.github.jruby-gradle.base' version '0.1.5'
     id 'java'
 }
+
 import com.github.jrubygradle.JRubyExec
 
 allprojects {
     group = 'org.embulk.input.jdbc'
     version = '0.6.0'
+
+    apply plugin: 'jacoco'
+    jacoco {
+        toolVersion = '0.7.1.201405082137'
+    }
 }
 
 subprojects {
     apply plugin: 'maven'  // install jar files to the local repo: $ gradle install
     apply plugin: 'java'
     //apply plugin: 'findbugs'
-    //apply plugin: 'jacoco'
     apply plugin: 'com.github.jruby-gradle.base'
 
     repositories {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -73,6 +89,27 @@ end
...
+    executionData = files(subprojects.jacocoTestReport.executionData)
+    reports {
+        html.enabled = true
+        xml.enabled = true
+        csv.enabled = false
+    }
+    onlyIf = {
+        true
+    }
+    doFirst {
```
frsyuki: Why is this necessary?

build.gradle
```diff
@@ -73,6 +89,27 @@ end
...
+    additionalSourceDirs = files(subprojects.sourceSets.main.allSource.srcDirs)
+    sourceDirectories = files(subprojects.sourceSets.main.allSource.srcDirs)
+    classDirectories =  files(subprojects.sourceSets.main.output)
+    executionData = files(subprojects.jacocoTestReport.executionData)
+    reports {
+        html.enabled = true
+        xml.enabled = true
+        csv.enabled = false
+    }
+    onlyIf = {
```
frsyuki: Why is this necessary?

build.gradle
```diff
@@ -4,18 +4,23 @@ plugins {
...
+
 import com.github.jrubygradle.JRubyExec
 
 allprojects {
     group = 'org.embulk.input.jdbc'
     version = '0.6.0'
+
+    apply plugin: 'jacoco'
+    jacoco {
+        toolVersion = '0.7.1.201405082137'
```
frsyuki: Is this necessary? Can't we use the latest version every time by removing this section?

build.gradle
```diff
@@ -33,6 +38,17 @@ subprojects {
         testCompile 'junit:junit:4.+'
     }
 
+    jacocoTestReport {
+        additionalSourceDirs = files(sourceSets.main.allSource.srcDirs)
+        sourceDirectories = files(sourceSets.main.allSource.srcDirs)
```
frsyuki: Are these lines necessary? According to code (https://github.com/gradle/gradle/blob/master/subprojects/jacoco/src/main/groovy/org/gradle/testing/jacoco/tasks/JacocoReport.groovy#L195 ), JacocoReport uses `allJava.getSrcDirs()` as the default setting for `sourceDirectories`.

PR #25, opened 2015-10-13
Add integration tests

muga:
Add integration tests for embulk-input-mysql, postgresql and redshift. Those integration tests are executed by `./gradlew test` command. They use actual database servers like host name, database name, user name and password. Those configurations are required as environment variables. If the variables are not set, the integration tests are skipped.


frsyuki:
Shouldn't we say this is a unit test? I think we can simply put all tests at `embulk-input-*/src/test/java/org/embulk/input` directory instead of `embulk-input-*/src/test/java/org/embulk/input/integration` directory.

frsyuki:
Looks well organized tests :+1: Code itself looks good.

muga:
git push -f

muga:
> Shouldn't we say this is a unit test? I think we can simply put all tests at embulk-input-*/src/test/java/org/embulk/input directory instead of embulk-input-*/src/test/java/org/embulk/input/integration directory.
Any thoughts?

Yes. Of course, we can say unit tests for Embulk plugins.

PR #26, opened 2015-11-06
Add some SSL options and update connector to latest.

kamatama41:
I want to use this plugin for [Heroku Postgres](https://www.heroku.com/postgres).
However, some SSL options are not implemented. ([ref](https://devcenter.heroku.com/articles/connecting-to-relational-databases-on-heroku-with-java#connecting-to-a-database-remotely))
In addition, current Postgres JDBC Driver has the [SSL issue](https://github.com/pgjdbc/pgjdbc/issues/259).

### DIFF
--- a/embulk-input-postgresql/README.md
+++ b/embulk-input-postgresql/README.md
@@ -18,6 +18,8 @@ PostgreSQL input plugins for Embulk loads records from PostgreSQL.
 - **fetch_rows**: number of rows to fetch one time (used for java.sql.Statement#setFetchSize) (integer, default: 10000)
 - **connect_timeout**: timeout for establishment of a database connection. (integer (seconds), default: 300)
 - **socket_timeout**: timeout for socket read operations. 0 means no timeout. (integer (seconds), default: 1800)
+- **ssl**: Connect using SSL? (boolean, default: false)
+- **sslfactory**: A class name to use as the SSLSocketFactory when establishing a SSL connection. [See also](https://jdbc.postgresql.org/documentation/head/ssl-factory.html) (string, default: null)
 - **options**: extra JDBC properties (hash, default: {})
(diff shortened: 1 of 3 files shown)


frsyuki:
I think setting sslfactory is not good idea because anyone will set org.postgresql.ssl.NonValidatingFactory. So as commented at `PostgreSQLOutputPlugin.getConnector` method, using ssl and sslverify options sounds better

frsyuki:
I created #27 (and https://github.com/embulk/embulk-output-jdbc/pull/65).

kamatama41:
OK, I agree with it.
Does It mean this issue is closed and #27 is applied?

kamatama41:
Incidentally, `props.setProperty("ssl", "true");` doesn't work at current Postgres JDBC driver because of [this issue](https://github.com/pgjdbc/pgjdbc/issues/259).
So #27 (and https://github.com/embulk/embulk-output-jdbc/pull/65) needs to update Driver version or include the option as url parameter (`postgresql://example.com:5432/some_db?ssl=true&sslfactory=~`).

frsyuki:
I updated JDBC driver version. Thank you!

hakobera:
@frsyuki I think postgresql JDBC driver version is not updated on this input plugin yet. https://github.com/embulk/embulk-input-jdbc/blob/master/embulk-input-postgresql/build.gradle#L4

You did it on [embulk-output-jdbc](https://github.com/embulk/embulk-output-jdbc/pull/65)

frsyuki:
oh, that's strange...github includes https://github.com/embulk/embulk-input-jdbc/pull/26/files#diff-ff437ce1902283bc4b10b297729334cdR4 but isn't applied. I apply it manually. thanks.

PR #32, opened 2016-01-09
Change timing to log SQL

hito4t:
The query SQL is logged at `run` method, but the SQL is used at `transaction` method before `run`.
So if the SQL fails at `transaction` method, we won't be able to know the SQL.
I've moved SQL logging from `run` method to `transaction` method.


frsyuki:
:+1:

PR #36, opened 2016-02-01
Support json type

joker1007:
This PR adds json type support.

- `string` can convert to `json`
- other type cannot convert to `json`, set `null`
- default type of PostgreSQL json(jsonb) column is `json`
  - get value as string from DB, and parse it.
- parse JSON string by JsonParser class of embulk-core

But I checked on only MySQL and PostgreSQL.
I cannot ready redshift, oracle, and sqlserver.
Would you confirm this PR on them?

### DIFF
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -166,7 +166,7 @@ public abstract class AbstractJdbcInputPlugin
         // validate column_options
         newColumnGetters(task, querySchema, null);
 
-        ColumnGetterFactory factory = new ColumnGetterFactory(null, task.getDefaultTimeZone());
+        ColumnGetterFactory factory = newColumnGetterFactory(null, task.getDefaultTimeZone());
         ImmutableList.Builder<Column> columns = ImmutableList.builder();
         for (int i = 0; i < querySchema.getCount(); i++) {
             JdbcColumn column = querySchema.getColumn(i);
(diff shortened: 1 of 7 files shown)


hito4t:
Thank you for the PR!

It will work well, however, I have some comments.


Only PostgreSQL has json type, and MySQL, Redshift, Oracle and SQL Server don't have json type.
So It is better to treat "json"/"jsonb" type name in embulk-input-postgresql than treat them in embulk-input-jdbc.
https://github.com/joker1007/embulk-input-jdbc/blob/support-json-type/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/ColumnGetterFactory.java#L126

You can modify `AbstractJdbcInputPlugin` class to enable subclasses to extend `ColumnGetterFactory` class as embulk-output-sqlserver extends `ColumnSetterFactory`.
https://github.com/embulk/embulk-output-jdbc/blob/master/embulk-output-sqlserver/src/main/java/org/embulk/output/SQLServerOutputPlugin.java#L151


`JsonParser` class seems to be thread-safe, but I think it is better to use it as non-static because `CsvParserPlugin` does.
https://github.com/embulk/embulk/blob/master/embulk-standards/src/main/java/org/embulk/standards/CsvParserPlugin.java#L234

joker1007:
Thanks for your advice!!
I will fix them. Please wait.  :muscle:

joker1007:
@hito4t 
I fixed it.
Please review me.

hito4t:
Thank you!

The column type name "json" and "jsonb" are peculiar for PostgreSQL, but the `valueType` "json" is common in all RDMSs.
(`valueType` represents a kind of `ColumnGetter`. Namely, the `valueType` "json" represents `JsonColumnGetter`.)

So `ColumnGetterFactory` (not `PostgreSQLColumnGetterFactory`) should create `JsonColumnGetter` if `valueType` is "json".
https://github.com/embulk/embulk-input-jdbc/pull/36/files#diff-e68d4b51ddae3f1b6d8267cb9d5b62bcR24

joker1007:
@hito4t 
Thanks!! it is mistake. :sweat_smile: 
I fixed it, and rebased.
How about is this?

hito4t:
Thank you!

PR #44, opened 2016-03-17
Feature convert date to string

yoyama:
This is an option which converts JDBC date type to embulk string with timestamp_format as default.


yoyama:
I would like to close this PR because this feature is covered by PR #45.

PR #45, opened 2016-03-18
Feature default column options

yoyama:
This 'default_column_options' supports column_options for each embulk type as default.


hito4t:
Thank you for the PR!
It must be very useful feature.

I have some comments as follows.
- I think if the key of `default_column_options` is value_type (sql type) instead of embulk type, this feature is more useful. For example, we can easily write yml for converting all decimal columns to string. And we can set different timestamp format for date column and datetime column. What do you think?
- In `AbstractJdbcInputPlugin#newColumnGetters`, `c_type` doesn't meet the Java naming standards. It should be `cType` or `columnType`.
- In `AbstractJdbcInputPlugin#columnOptionOf`, `a.or(b).or(c)` is slightly readable than `a.or(b.or(c))`.
- `ColumnGetterFactory#defaultColumnOptions` is not used.

yoyama:
Changed the key from value type to JDBC type as result of discussion.

hito4t:
Thanks!

But these don't seem to be fixed yet. Please verify.
- In `AbstractJdbcInputPlugin#columnOptionOf`, a.or(b).or(c) is slightly readable than a.or(b.or(c)).
- `ColumnGetterFactory#defaultColumnOptions` is not used.

yoyama:
Thank you for your reviews.
I fixed them. 
Could you review again?

hito4t:
Thank you!

PR #50, opened 2016-04-30
Running after_select after pageBuilder.finish

frsyuki:
`after_select` is used to store last status of a bulk loading. If an
error happened after `after_select` before committing data to the output
storage, it causes inconsistency. Output data is not stored on the
storage but `after_select` is already executed.

To make it less likely, this change runs `after_select` after
`pageBuilder.finish` because some output plugins do significant works
in `finish` that may throw exceptions.


muga:
This PR could reduce data inconsistency issue:-) Anyone merge this into master? and release new version? or any comment?

PR #51, opened 2016-05-27
Ignore case in table name and column names

hito4t:
Because embulk-output-jdbc will ignore case in table name and column names (https://github.com/embulk/embulk-output-jdbc/pull/108/files), embulk-input-jdbc should also ignore case.


embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
```diff
@@ -146,25 +152,90 @@ protected String quoteIdentifierString(String str)
...
     protected String buildTableName(String tableName)
     {
-    	return quoteIdentifierString(tableName);
+        return quoteIdentifierString(tableName);
     }
 
     public String buildSelectQuery(String tableName,
             Optional<String> selectColumnList, Optional<String> whereCondition,
-            Optional<String> orderByColumn)
+            Optional<String> orderByColumn) throws SQLException
```
frsyuki: orderByColumn is not a column name. It is an expression including DESC or ASC options.

PR #54, opened 2016-07-13
Add jdbcopts for Redshift

yu-yamada:
Add Jdbcopts option.
Especially for SSL.


hito4t:
Thank you for the PR!

Because PostgreSQLInputPlugin already supported `ssl` property, we should add the same property to RedshiftInputPlugin for SSL.
https://github.com/embulk/embulk-input-jdbc/pull/27/files

Although JdbcInputPlugin already has `options` property for JDBC options, a property specialized for SSL should be useful.

yu-yamada:
Thank you for your reply.
I had overlooked the setting of PostgreSQL.
So, I changed same as PostgreSQL :-P
It work.

PR #55, opened 2016-07-29
[WIP][Proposal] Support incremental load

sakama:
I implemented **incremental load** as new feature.
This feature allows us to load record incrementally.

How do you think about this feature?

Currentlly I checked only with MySQL. If you think good to implement,  I'll check with other RDBMS.

After this PR will be merged, I'll also take over [Add unit tests #25] to add unit tests.

## config.yml and example data
I added 2 options, `incremental_column` and `last_record`.
Both of `incremental_column` and `last_record` accepts **multiple values**.

These options works with `table`, `select`, `where`, `order_by` options, but doesn't works with `query` option.
I think it's difficult to parse SQL query and rebuild new query.

```yaml
# config.yml
in:
  type: mysql
  host: localhost
  user: username
  password: password
  database: incremental
  table: test
  select: "*"
  where: name IS NOT NULL
  order_by:
  incremental_column:
    - "time"
    - "id"
  last_record:
out:
  type: stdout
```

```sh
# data
mysql> SELECT * FROM `test` ORDER BY `time` ASC, `id` ASC
+----+--------------+---------------------+
| id | name         | time                |
+----+--------------+---------------------+
|  5 | pqrstuv      | 2016-01-23 13:23:26 |
|  4 | hijklmno     | 2016-06-23 13:23:26 |
|  3 | abcdefg      | 2016-07-29 14:29:50 |
|  2 | embulk jruby | 2016-07-29 14:56:35 |
|  1 | embulk       | 2016-07-29 14:59:32 |
+----+--------------+---------------------+
```

### First execution

```sh
$ embulk run config.yml -c diff.yml
```

Then, plugin automatically add `ORDER BY` clause and build following query.

```sql
SELECT * FROM `test` WHERE name IS NOT NULL ORDER BY `time` ASC, `id` ASC
```

Finally, plugin generates diff.yml with `last_record` value.
```yaml
# diff.yml
in:
  last_record: {time: '2016-07-29 14:59:32.0', id: '1'}
out: {}
```

### Second execution

At second execution, plugin uses values that was written at diff.yml

```sh
$ embulk run config.yml -c diff.yml
```

Then, plugin will automatically add `WHERE` and `ORDER_BY` clause and build following query.
If new records exists at RDBMS, **we can load only new records**.

```sql
SELECT * FROM `test` WHERE time > "2016-07-29 14:59:32.0" AND id > "1" AND id IS NOT NULL ORDER BY `time` ASC, `id` ASC
```

### DIFF
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -1,5 +1,6 @@
 package org.embulk.input.jdbc;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.nio.file.Paths;
@@ -66,10 +67,14 @@ public abstract class AbstractJdbcInputPlugin
         @ConfigDefault("null")
         public Optional<String> getOrderBy();
 
-        //// TODO See bellow.
-        //@Config("last_value")
-        //@ConfigDefault("null")
-        //public Optional<String> getLastValue();
+        @Config("incremental_column")
+        @ConfigDefault("null")
+        public Optional<List<String>> getIncrementalColumn();
+
+        @Config("last_record")
+        @ConfigDefault("null")
+        public Optional<Map<String, String>> getLastRecord();
+        public void setLastRecord(Optional<Map<String, String>> lastRecord);
 
         // TODO limit_value is necessary to make sure repeated bulk load transactions
         //      don't a same record twice or miss records when the column
@@ -149,10 +154,6 @@ public abstract class AbstractJdbcInputPlugin
     {
         PluginTask task = config.loadConfig(getTaskClass());
 
-        //if (task.getLastValue().isPresent() && !task.getOrderBy().isPresent()) {
-        //    throw new ConfigException("order_by parameter must be set if last_value parameter is set");
-        //}
-
         Schema schema;
         try (JdbcInputConnection con = newConnection(task)) {
             schema = setupTask(con, task);
@@ -192,11 +193,13 @@ public abstract class AbstractJdbcInputPlugin
             if (task.getTable().isPresent() || task.getSelect().isPresent() ||
                     task.getWhere().isPresent() || task.getOrderBy().isPresent()) {
                 throw new ConfigException("'table', 'select', 'where' and 'order_by' parameters are unnecessary if 'query' parameter is set.");
+            } else if (task.getIncrementalColumn().isPresent() || task.getLastRecord().isPresent()) {
+                throw new ConfigException("'incremental_column', 'last_record' parameters couldn't be used if 'query' parameter is set.");
             }
             return task.getQuery().get();
         } else if (task.getTable().isPresent()) {
             return con.buildSelectQuery(task.getTable().get(), task.getSelect(),
-                    task.getWhere(), task.getOrderBy());
+                    task.getWhere(), task.getOrderBy(), task.getIncrementalColumn(), task.getLastRecord());
         } else {
             throw new ConfigException("'table' parameter is required (if 'query' parameter is not set)");
         }
@@ -224,12 +227,9 @@ public abstract class AbstractJdbcInputPlugin
     protected ConfigDiff buildNextConfigDiff(PluginTask task, List<TaskReport> reports)
     {
         ConfigDiff next = Exec.newConfigDiff();
-        // TODO
-        //if (task.getOrderBy().isPresent()) {
-        //    // TODO when parallel execution is implemented, calculate the max last_value
-        //    //      from the all commit reports.
-        //    next.set("last_value", reports.get(0).get(JsonNode.class, "last_value"));
-        //}
+        if (reports.size() > 0 && reports.get(0).has("last_record")) {
+            next.set("last_record", reports.get(0).get(Map.class, "last_record"));
+        }
         return next;
     }
 
@@ -258,7 +258,7 @@ public abstract class AbstractJdbcInputPlugin
                 while (true) {
                     // TODO run fetch() in another thread asynchronously
                     // TODO retry fetch() if it failed (maybe order_by is required and unique_column(s) option is also required)
-                    boolean cont = fetch(cursor, getters, pageBuilder);
+                    boolean cont = fetch(cursor, getters, pageBuilder, task);
                     if (!cont) {
                         break;
                     }
@@ -284,10 +284,9 @@ public abstract class AbstractJdbcInputPlugin
         }
 
         TaskReport report = Exec.newTaskReport();
-        // TODO
-        //if (orderByColumn != null) {
-        //    report.set("last_value", lastValue);
-        //}
+        if (task.getIncrementalColumn() != null) {
+            report.set("last_record", task.getLastRecord().get());
+        }
         return report;
     }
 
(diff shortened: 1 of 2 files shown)


embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -66,10 +67,14 @@
         @ConfigDefault("null")
         public Optional<String> getOrderBy();
 
-        //// TODO See bellow.
-        //@Config("last_value")
-        //@ConfigDefault("null")
-        //public Optional<String> getLastValue();
+        @Config("incremental_column")
```
frsyuki: should be incremental_column"s"

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
```diff
@@ -187,35 +190,69 @@ public String buildSelectQuery(String tableName,
...
+            if (incrementalColumn.isPresent()) {
+                for (int i = 0; i < incrementalColumn.get().size(); i++) {
+                    String column = incrementalColumn.get().get(i);
+                    sb.append(quoteIdentifierString(column)).append(" ASC");
+                    if (i < incrementalColumn.get().size() - 1) {
+                        sb.append(", ");
+                    }
+                }
+            }
+            if (orderByColumn.isPresent() && actualOrderByColumn != null) {
```
frsyuki: Sorting records using `order_by` here doesn't make sense because records in the results are not actually sorted by `order_by` option. Records are sorted by `incremental_columns`. `order_by` option is ignored from user's point of view.

frsyuki: it sounds better to make it error if both of `order_by` and `incremental_columns` are set.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -66,10 +67,14 @@
...
+        @ConfigDefault("null")
+        public Optional<List<String>> getIncrementalColumn();
```
frsyuki: incremental_column option needs validation.

* all of incremental_columns must be included in the selected fields.
* if incremental, setting order_by is invalid. it should be rejected.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -362,6 +367,16 @@ private boolean fetch(BatchSelect cursor,
                 reportRows *= 2;
             }
         } while (result.next());
+
+        if (task.getIncrementalColumn().isPresent()) {
+            for (String key : task.getIncrementalColumn().get()) {
+                if (temporaryLastRecord.containsKey(key)) {
```
frsyuki: this validation logic seems invalid because incremental loading doesn't work as expected if the results don't include all of the incremental_column columns.

frsyuki: This seems invalid because column names are case insensitive in SQL. If column names of the target table don't match with selecting column names, column names in the results may not match with selecting column names.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -340,20 +339,26 @@ public JdbcColumnOption get()
...
         List<Column> columns = pageBuilder.getSchema().getColumns();
         long rows = 0;
         long reportRows = 500;
         do {
+            temporaryLastRecord.clear();
             for (int i=0; i < getters.size(); i++) {
                 int index = i + 1;  // JDBC column index begins from 1
                 getters.get(i).getAndSet(result, index, columns.get(i));
+                String k = columns.get(i).getName();
+                temporaryLastRecord.put(k, result.getString(index));
```
frsyuki: Using Java's Map with String key is invalid because column names are case insensitive in SQL while String is case sensitive. If column names of the target table don't match with selecting column names, column names in the results may not match with selecting column names.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -284,10 +284,9 @@ public TaskReport run(TaskSource taskSource,
         }
 
         TaskReport report = Exec.newTaskReport();
-        // TODO
-        //if (orderByColumn != null) {
-        //    report.set("last_value", lastValue);
-        //}
+        if (task.getIncrementalColumn() != null) {
+            report.set("last_record", task.getLastRecord().get());
```
frsyuki: last_record will be invalid value if this bulk data loading loaded 0 records (that may happen as a successful case).

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -224,12 +227,9 @@ public ConfigDiff guess(ConfigSource config)
...
         ConfigDiff next = Exec.newConfigDiff();
-        // TODO
-        //if (task.getOrderBy().isPresent()) {
-        //    // TODO when parallel execution is implemented, calculate the max last_value
-        //    //      from the all commit reports.
-        //    next.set("last_value", reports.get(0).get(JsonNode.class, "last_value"));
-        //}
+        if (reports.size() > 0 && reports.get(0).has("last_record")) {
+            next.set("last_record", reports.get(0).get(Map.class, "last_record"));
+        }
```
frsyuki: last_record is lost if this bulk data loading loaded 0 records (that may happen as a successful case).

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
```diff
@@ -187,35 +190,69 @@ public String buildSelectQuery(String tableName,
...
         sb.append(" FROM ").append(buildTableName(actualTableName));
 
-        if (whereCondition.isPresent()) {
-            sb.append(" WHERE ").append(whereCondition.get());
+        if (whereCondition.isPresent() || incrementalColumn.isPresent()) {
+            sb.append(" WHERE ");
+            if (incrementalColumn.isPresent() && lastRecord.isPresent()) {
+                for (int i = 0; i < incrementalColumn.get().size(); i++) {
+                    String column = incrementalColumn.get().get(i);
+                    sb.append(column).append( " > \"");
```
frsyuki: this quoting logic of a last value is invalid. Using `"` for string literal is not valid as a SQL standard (it's a MySQL extension). It should be `'`.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
```diff
@@ -187,35 +190,69 @@ public String buildSelectQuery(String tableName,
...
+                    sb.append(lastRecord.get().get(column).replace("\"", "\\\""));
```
frsyuki: Escaping quoting character with `\` is invalid as a standard SQL (it's a MySQL extension). It doesn't work with PostgreSQL, for example.

hito4t:
In the example, a newer record has a smaller id.
If so, a next new record will be like `{id: 0, time: 2016-08-01 09:00:00}`.
But the record won't be selected because the record doesn't meet the condition `id > 1`.

If the example is wrong and `id` always increases, we should use only `id` as `incremental_column`.

And if a table has composite key, it won't work as expected.
For example, if there are records as follows,
<pre>
+----+--------+
| id | sub_id |
+----+--------+
|  1 |      1 |
|  1 |      2 |
|  1 |      3 |
|  2 |      1 |
|  2 |      2 |
|  2 |      3 |
+----+--------+
</pre>
the condition for next records will be `id > 2 AND sub_id > 3`.
But a new record `{id: 3, sub_id: 1}` won't be selected.
The condition should be `id = 2 AND sub_id > 3 OR id > 2`.

frsyuki:
@hito4t that's a very good point. code should be fixed.

frsyuki:
Fixed at #55.

sakama:
@frsyuki Thank you for your work.
@hito4t Thanks your comment.

Implementation and discussion is now progressing at #57. So I close this PR.

PR #56, opened 2016-07-29
Allow DESC order at order_by option

frsyuki:
As like other query-part options (`select` and `where`), `order_by` option can assume that the given parameter is an expression. A benefit is that order_by can include multiple columns (e.g. `category, created_at` and/or DESC option `category, created_at DESC`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,11 @@ See [embulk-input-sqlserver](embulk-input-sqlserver/).
   - **query**: SQL to run (string)
 - If **query** is not set,
   - **table**: destination table name (string, required)
-  - **select**: comma-separated list of columns to select (string, default: "*")
+  - **select**: expression of select (e.g. `id, created_at`) (string, default: "*")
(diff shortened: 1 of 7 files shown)


hito4t:
That's a good idea.

The table name will be quoted by embulk-output-jdbc, and other expressions (select, where, order_by) will be used as they are.
I think it's a bit confusing.

If we add `order_by` to the first example of README, and show the SQL which embulk-output-jdbc will generate, users can understand easily.

<pre>
  ...
  table: "my_table"
  select: "col1, col2, col3"
  where: "col4 != 'a'"
  order_by: "col1 DESC"

The following SQL will be generated.
SELECT col1, col2, col3 FROM `my_table` WHERE col4 != 'a' ORDER BY col1 DESC
</pre>

frsyuki:
good point. I'll add

frsyuki:
added

hito4t:
Thank you!

PR #57, opened 2016-07-30
Support incremental load

frsyuki:
Design review of #55.
This assumes that #56 is merged.

This code supports only integers as strings as last_record option. It needs to be strict typing to support postgresql which doesn't have much implicit type conversion. Implementation would use ColumnGetter mechanism for type conversion from ResultSet to `Map<String, Object> lastRecord`, and from `Map<String, Object> lastRecord` to `PreparedStatement.setXxx(index, thisValue)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ See [embulk-input-sqlserver](embulk-input-sqlserver/).
 
 Incremental loading uses monotonically increasing unique columns (such as auto-increment id) to load records inserted (or updated) after last execution.
 
-First, if `increment: true` is set, this plugin loads all records with additional ORDER BY. For example, if `incremental_columns: [updated_at, id]` option is set, query will be as following:
+First, if `incremental: true` is set, this plugin loads all records with additional ORDER BY. For example, if `incremental_columns: [updated_at, id]` option is set, query will be as following:
 
 ```
 SELECT * FROM (
--- a/README.md
+++ b/README.md
@@ -64,9 +64,9 @@ See [embulk-input-sqlserver](embulk-input-sqlserver/).
 
 ### Incremental loading
 
-If `incremental: true` option is set, incremental loading is enabled. Incremental loading generates additional WHERE conditions on one multiple incrementing columns (such as auto-increment id) to load records inserted (or updated) after last execution.
+Incremental loading uses monotonically increasing unique columns (such as auto-increment id) to load records inserted (or updated) after last execution.
 
-If `increment: true` is set, this plugin loads all records with additional ORDER BY. For example, if `incremental_columns: [updated_at, id]` option is set, query will be as following:
+First, if `increment: true` is set, this plugin loads all records with additional ORDER BY. For example, if `incremental_columns: [updated_at, id]` option is set, query will be as following:
 
 ```
 SELECT * FROM (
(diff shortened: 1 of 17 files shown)


frsyuki:
@sakama I implement my comments at #55. Could you take over?

sakama:
@frsyuki Thank you for your works.
I added small fixes and trying to support other types(especially time, timestamp, date types).

frsyuki:
@sakama I think we don't have to support timestamp types at the first release. we can add them later if someone needs.

sakama:
@frsyuki Thank you for your works.
I  tried with following RDBMS and confirmed this implementation works fine.
* PostgreSQL 9.4.4
* MySQL 5.7.11
* Microsoft SQL Server 12.00.811 (Microsoft Azure SQL Database)
* AWS Redshift

> I think we don't have to support timestamp types at the first release. we can add them later if someone needs.

I got it.
I'll send another PR when I need other types supports.

hito4t:
Would you check the following comments?

- `AbstractJdbcInputPlugin.PluginTask` has `builtQuery:String` and `builtQueryParameters:List<JdbcLiteral>` separately, but they are closely related, and actually there is a class `JdbcInputConnection.PreparedQuery` which has both properties. Can we replace the two properties of  `AbstractJdbcInputPlugin.PluginTask` with one property whose type is `PreparedQuery`?

- LastRecord is set in `AbstractJdbcInputPlugin#fetch`, but `fetch` is called in a loop (https://github.com/embulk/embulk-input-jdbc/blob/ad8301ee5eca2bb0deedf3b8c24a04dc81dfe370/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java#L428). Namely, lastRecord might be set multiple times. LastRecord should be set out of the loop.

- Can you write README for the feature?

frsyuki:
@hito4t Thank you for your comments :-)

* Sounds good. I made PreparedQuery serializable.

* It is a bug :( I fixed it.

* I added some docs. Please check them.

hito4t:
Thank you!
I've checked the latest commits.

- `JdbcInputConnection#newSelectCursor` and `newBatchSelect` take `String` and `List<JdbcLiteral>` as parameters, but it is simpler if they take only `PreparedQuery`  as a parameter.
- I feel the implementation about LastRecordStore is a bit redundant. How about following implementation?

1. remove this if statement.
https://github.com/embulk/embulk-input-jdbc/blob/10ed219b3c8ba5c345e715c98e1a144aad08ad6d/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java#L416

2. https://github.com/embulk/embulk-input-jdbc/blob/10ed219b3c8ba5c345e715c98e1a144aad08ad6d/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java#L432
<pre>
            if (lastRecordStore != null && totalRows > 0) {
                lastRecordStore.accept(getters);
            }
</pre>
↓
<pre>
            if (task.getIncremental() && totalRows > 0) {
                lastRecordStore = new LastRecordStore(task.getIncrementalColumnIndexes(), task.getIncrementalColumns());
                lastRecordStore.accept(getters);
            }
</pre>

3. https://github.com/embulk/embulk-input-jdbc/blob/10ed219b3c8ba5c345e715c98e1a144aad08ad6d/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java#L456
<pre>
 if (lastRecordStore != null && totalRows > 0) {
</pre>
↓
<pre>
 if (lastRecordStore != null) {
</pre>

- `increment: true` should be `incremental: true`. https://github.com/embulk/embulk-input-jdbc/pull/57/files#diff-04c6e90faac2675aa89e2176d2eec7d8R69

frsyuki:
> JdbcInputConnection#newSelectCursor

👍 

> increment: true should be incremental: true

Good catch 👍  Fixed.

frsyuki:
> LastRecordStore

sounds good. I fixed accordingly.
You can comment in-line on the code, by the way.

PR #61, opened 2016-08-19
postgresql: support application_name option

frsyuki:
This is useful for system administrators to know that a long-running query is issued using embulk-input-postgresql plugin.


hito4t:
https://jdbc.postgresql.org/documentation/head/connect.html

PR #65, opened 2016-09-08
embulk-input-db2

hito4t:
plugin for IBM DB2


frsyuki:
Great!

frsyuki:
Code looks good.
A problem is that I know almost nothing about DB2...so I believe you. Please merge if you think it's good (could you add [me](https://rubygems.org/profiles/frsyuki) to an owner when you release a gem? Thank you!).

PR #66, opened 2016-09-16
Fix thinko in normalizeTableNameCase().

kasaharatt:
At the second con.tableExists() in normalizeTableNameCase(), the argment should be lowerTablename, not upperTableName.

regards,


hito4t:
Thank you!

PR #75, opened 2016-11-02
[WIP] Support timestamp/timestamptz type for PostgreSQL incremental_columns

sakama:
Revise of #73.

This pull-requests focused to PostgreSQL. Additionally supported `timestamp` and `timestamptz`(timestamp with timezone) type columns as incremental_columns.

Plugin will raise exception when...
* Column type is timestamp and DB type is not PostgreSQL
* Column type is time, timetz, or any types that don't include year/month/date
* Column type is date, or any types that don't include hour/minute/seconds

### DIFF
--- a/embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/getter/TimestampColumnGetter.java
+++ b/embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/getter/TimestampColumnGetter.java
@@ -1,0 +1,78 @@
+package org.embulk.input.postgresql.getter;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import org.embulk.config.ConfigException;
+import org.embulk.input.jdbc.getter.AbstractTimestampColumnGetter;
+import org.embulk.spi.PageBuilder;
+import org.embulk.spi.time.Timestamp;
+import org.embulk.spi.time.TimestampFormatter;
+import org.embulk.spi.type.Type;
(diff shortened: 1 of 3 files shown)


embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/getter/TimestampColumnGetter.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    public void decodeFromJsonTo(PreparedStatement toStatement, int toIndex, JsonNode fromValue)
+            throws SQLException
+    {
+        switch (columnTypeName) {
+            case "timestamp":
+                toStatement.setTimestamp(toIndex, java.sql.Timestamp.valueOf(fromValue.asText()));
+                break;
+            case "timestamptz":
+                SimpleDateFormat dateFormatter = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSSSSS z");
```
sakama: I want to use `org.embulk.spi.time.TimestampParser` instead of SimpleDateFormat.
Currently TimestampParser throws `TimestampParseException` and I'm investigating.

frsyuki:
Apparently, some code still don't work at all. I added tests on #77 (`support-timestamp-for-incremental-with-test` branch). How about using that branch to fix some behavior?

sakama:
Thanks. I close this pull-request.

PR #76, opened 2016-11-02
Adding test cases based on embulk-test

frsyuki:
This adds test cases using `embulk-test` package: https://github.com/embulk/embulk/pull/503
This PR includes temporary changes to use the pre-release version of embulk. They should be removed before merging this change.
Travis-CI won't pass until new version of embulk is released with embulk-test.

PR #78, opened 2016-11-03
Use custom column getter of hstore only when to_type is specifically set to json

frsyuki:
This is a refactoring.
This assumes #76. Diff from #76 is https://github.com/embulk/embulk-input-jdbc/compare/postgres-test...hstore-json-ext


muga:
This PR is closed. This commits will be merged on #79.

PR #79, opened 2016-11-03
Adding support for incremental loading using timestamp and timestamptz types to PostgreSQL

frsyuki:
This PR assumes #78. Diff from #78 is https://github.com/embulk/embulk-input-jdbc/compare/hstore-json-ext...incremental-timestamp

PR #80, opened 2016-11-04
Use embulk v0.8.15 for incremental timestamp

muga:
Upgrade Embulk version to v0.8.15 for #79.


sakama:
Fixed CI failure.
I also tried to add MySQL CI setting but still failing.
I'll take care at another pull-req #82.

PR #83, opened 2016-11-10
[WIP] Adding support for incremental loading using datetime and timestamp types to MySQL

sakama:
Similar with PostgreSQL timestamp/timestamptz incremental load supports #79.

This implementation supports `TIMESTAMP` and `DATETIME` type columns of MySQL.

I don't rewrite existing test case with new EmbulkTest framework for now.
Current test case of embulk-input-mysql is using CSV formatter and EmbulkTest only accepts inputConfig, doen't accept outputConfig.
https://github.com/embulk/embulk/blob/master/embulk-test/src/main/java/org/embulk/test/TestingEmbulk.java#L170

### DIFF
--- a/embulk-input-mysql/README.md
+++ b/embulk-input-mysql/README.md
@@ -150,3 +150,17 @@ in:
 ```
 $ ./gradlew gem
 ```
+
+Running tests:
+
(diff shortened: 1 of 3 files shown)


sakama:
This feature will be handled at #89. I close this issue

PR #85, opened 2016-11-18
Apply test jvmArgs and environment to postgresql project only

muga:
I noticed that the commit https://github.com/embulk/embulk-input-jdbc/commit/b37ea5c361477ec7d86658192e4c2ec69b9e4dac broke unit tests for each projects except postgresql's. This settings is good for unit tests running on Travis-CI. But, now it's better to apply postgresql test only.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -102,10 +102,15 @@ subprojects {
         }
     }
 
-    test {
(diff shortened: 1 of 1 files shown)


muga:
@hito4t I merge this because the current settings has broken unit tests on your local.

PR #89, opened 2016-11-22
MySQL incremental loading supports DateTime and Timestamp types

muga:
This PR is follow up of #83 

I implemented TIMESTAMP and DATETIME types support for MySQL incremental data loading and the unit tests.

To simplify this feature, users currently can use this feature if users configured `useLegacyDatetimeCode=false`. Because MySQL JDBC Driver's time_zone retrieval is complicated by the flag. Since next version of MySQL JDBC Driver (v6) will, by default, tune off the useLegacy, I think that it should be fine.


muga:
I put one more commit, that the plugin manually loads timezone mappings to com.mysql.jdbc.TimeUtil.timeZoneMappings field. The loading often fails because the specific properties file cannot be found by the Embulk parent classloader. 

I think that MySQL JDBC Driver should fix this issue but, I put the workaround here.

frsyuki:
LGTM 👍

frsyuki:
The comments on the `loadTimeZoneMappings` method is hard to understand why it's actually necessary or who is the cause of the workaround. How about saying like this?:

```
        // Here initializes com.mysql.jdbc.TimeUtil.timeZoneMappings static field by calling
        // static timeZoneMappings method using reflection.
        // The field is usually initialized when Driver#connect method is called. But the field
        // initialization fails when a) useLegacyDatetimeCode=false is set AND b) mysql server's
        // default_time_zone is not SYSTEM (default). According to the stacktrace, that's because
        // the com.mysql.jdbc.TimeUtil.loadTimeZoneMappings can't find TimeZoneMapping.properties
        // from the classloader. It seems like a bug of JDBC Driver where it should use the class loader
        // that loaded com.mysql.jdbc.TimeUtil class rather than system class loader to read the
        // property file because the file should be in the same classpath with the class.
        // Here implements a workaround as as workaround.
```

muga:
Fixed it. thank you for comments.

muga:
Merged cc: @frsyuki @hito4t

PR #90, opened 2016-11-24
wuzimei :: add blob(snappy uncompression) to string

wuzimei:
Hi, our input has a blob column in which we save a string value compressed by snappy. So I want to use embulk-input-jdbc to read uncompressed string value from blob column. 

I worked base on v0.8.0 embulk-input-jdbc. And I have not touched gradle before.

If possible, you can add the function by yourself due to my poor programming skill.

What I have done is as follow: 

I tried to make a SnappyColumnGetter similar to StringColumnGetter just add "Snappy.uncompressString" when fetch ResultSet value.

So I have succeeded to get snappyed blob value from our db by using config as following:
column_options:
    data: {value_type: snappy}

Thank you.


frsyuki:
I think filter plugin works better for this use case in terms of concept of Embulk. The biggest blocker is binary type support of Embulk: https://github.com/embulk/embulk/issues/420

PR #93, opened 2017-02-03
Set time zone explicitly in tests

hito4t:
in order not to depend on the time zone of test machine.


embulk-input-mysql/src/test/java/org/embulk/input/mysql/BasicTest.java
```diff
@@ -41,6 +42,7 @@ private static String readResource(String fileName)
     @Before
     public void setup()
     {
+        TimeZone.setDefault(TimeZone.getTimeZone("Europe/Helsinki")); // +200
```
muga: TravisCI has been using `gradle check` command. The gradle reads the following test settings that configure the time zone. Doesn't the setting work for you?
https://github.com/embulk/embulk-input-jdbc/blob/master/build.gradle#L110

muga: I was missing the settings included in https://github.com/embulk/embulk-input-jdbc/pull/94 and https://github.com/embulk/embulk-input-jdbc/pull/95.

frsyuki:
LGTM 👍

embulk-input-mysql/src/test/java/org/embulk/input/mysql/BasicTest.java
frsyuki: Right. This should be unnecessary because gradle sets it. #94 and #95 should also don't need this line. Other than that, LGTM 👍

hito4t: In my environment, `jvmArgs "-Duser.country=FI -Duser.timezone=Europe/Helsinki"` doesn't work.
Instead `jvmArgs "-Duser.country=FI", "-Duser.timezone=Europe/Helsinki"` works.
Is it because my gradle (2.12) is old?

hito4t: The gradle version is written in embulk-input-jdbc/gradle/wrapper/gradle-wrapper.properties

hito4t: `jvmArgs "-Duser.country=FI -Duser.timezone=Europe/Helsinki"` may not work in any environment.
But, in Linux `environment "TZ", "Europe/Helsinki"` may work and it doesn't work in Windows.

hito4t:
I've removed timezone setting from source codes and corrected timezone setting in `build.bradle` .

muga:
LGTM 👍  The fix works for me.

PR #98, opened 2017-02-10
Fix bug that prevents using non-standard SQL type even if value_type is set

hito4t:
For example, cannot use Oracle BINARY_DOUBLE type even if `value_type` of `column_options` is set.


sutoh:
👍

PR #101, opened 2017-04-05
Support Redshift timestamp/timestamptz incremental load

sakama:
Current version of embulk-input-redshift(v0.8.2) supports incremental load for `string` and `integer` columns.
However, doesn't supports `timestamp` `timestamptz` columns for now.
I added support for these type of columns based on implementation of PostgreSQL.

I've not changed .travis.yml because Travis doesn't support Redshift:).
Then, all test case for Redshift will be skipped.

~I'm checking more detail and will remove[WIP] after that.~


sakama:
Also updated README.
Incremental load example for timestamp column is already exists at [there](https://github.com/embulk/embulk-input-jdbc/tree/master/embulk-input-redshift#incremental-loading).
So I only updated about unit test.

hito4t:
Thank you for the PR!
I'll review it as soon as possible.

hito4t:
I've tried it and it worked fine!

By the way, I have a request for tests.
The default fetch size of embulk-input-jdbc is 10000.
But the maximum fetch size for single-node Redshift is 1000.
http://docs.aws.amazon.com/redshift/latest/dg/fetch.html
If fetch size in tests is set to 1000, developers can test easier.

sakama:
@hito4t Thank you for confirming.

> If fetch size in tests is set to 1000, developers can test easier.

I've already added `fetch_rows: 1000` at [base config](https://github.com/embulk/embulk-input-jdbc/pull/101/files#diff-cf7f885e8e1761d4b0a68737d0df200aR7).

Then, If developer's config is based on this config, base config will be loaded at [RedshiftTests.baseConfig()](https://github.com/embulk/embulk-input-jdbc/pull/101/files#diff-d9789d1d9e5a176e95c5e959c9b9fc77R16) and IncrementalTest class will [also load this base config](https://github.com/embulk/embulk-input-jdbc/pull/101/files#diff-17e6e31e7e93dcae80253a5dea65b4e2R45).

Something went wrong?

hito4t:
@sakama 

I'm sorry.
I missed `fetch_rows` when creating base config for my environment...

sakama:
thanks!

PR #108, opened 2017-05-26
Add Ssl enum

hiroyuki-sato:
Add enum Ssl for support SSL option (ex. embulk-input-mysql).

PR #106, opened 2017-04-29
The MySQL plugin compare server and client timezone

hiroyuki-sato:
@muga @hito4t 

Reference implementation [This comment](https://github.com/embulk/embulk-input-jdbc/issues/103#issuecomment-297924677).

The embulk-input-mysql to compare server and client timezone. 

I would like to request comments about this changes.

### Questions

* How to check server timezone parameter before load data. 
  * I made `before_load` method for executing timezone check SQL. 
* How to get server timezone parameter
  * I executed the query `select @@system_time_zone` for get server timezone. 
  * Older MySQL version doesn't support `@@system_time_zone` parameter. (I'll comment later)
  * I tried to use `getServerTimezoneTZ()` method. But It always returns `null`
* How to get the client side parameter. 
  * I used `System.getProperty("user.timezone");`
* How to compare client and server timezone. 
  * I used `sys_tz.hasSameRules(usr_tz)`. 
  * This method return `false` when I compare `GMT+9` and `JST` (See below)

### Same timezone 

```
2017-04-29 21:57:47.042 +0900: Embulk v0.8.18
2017-04-29 21:57:48.972 +0900 [INFO] (0001:preview): Loaded plugin embulk/input/mysql from a load path
2017-04-29 21:57:49.031 +0900 [INFO] (0001:preview): Fetch size is 10000. Using server-side prepared statement.
2017-04-29 21:57:49.446 +0900 [INFO] (0001:preview): Fetch size is 10000. Using server-side prepared statement.
2017-04-29 21:57:49.455 +0900 [WARN] (0001:preview): The plugin will set `useLegacyDatetimeCode=false` by default in future.
2017-04-29 21:57:49.456 +0900 [INFO] (0001:preview): SQL: SELECT * FROM `time_test`
2017-04-29 21:57:49.461 +0900 [INFO] (0001:preview): > 0.00 seconds
```

### Difference timezone

```
2017-04-29 22:00:55.618 +0900 [INFO] (0001:preview): Loaded plugin embulk/input/mysql from a load path
2017-04-29 22:00:55.671 +0900 [INFO] (0001:preview): Fetch size is 10000. Using server-side prepared statement.
2017-04-29 22:00:56.103 +0900 [INFO] (0001:preview): Fetch size is 10000. Using server-side prepared statement.
2017-04-29 22:00:56.119 +0900 [WARN] (0001:preview): The server timezone and client timezone are different, the plugin will fetch wrong datetime values.
[...]

PR #109, opened 2017-05-26
Add mysql ssl option

hiroyuki-sato:
The embulk-input-mysql support SSL connection. 
This PR require PR #108.

@hito4t Please take a look when you can.

### DIFF
--- a/embulk-input-jdbc/README.md
+++ b/embulk-input-jdbc/README.md
@@ -14,6 +14,7 @@ Generic JDBC input plugin for Embulk loads records from a database using a JDBC 
 - **url**: URL of the JDBC connection (e.g. 'jdbc:sqlite:mydb.sqlite3') (string, required)
 - **user**: database login user name (string, optional)
 - **password**: database login password (string, default: optional)
+- **ssl**: use SSL to connect to the database (string, default: `disable`. `enable` uses SSL without server-side validation and `verify` checks the certificate. For compatibility reasons, `true` behaves as `enable` and `false` behaves as `disable`.)
 - **schema**: destination schema name (string, default: use the default schema)
 - **fetch_rows**: number of rows to fetch one time (integer, default: 10000)
 - **connect_timeout**: not supported.
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/Ssl.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/Ssl.java
@@ -1,0 +1,37 @@
+package org.embulk.input.jdbc;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+import org.embulk.config.ConfigException;
+
+public enum Ssl
+{
+    ENABLE,
+    DISABLE,
+    VERIFY;
+
+    @JsonValue
+    @Override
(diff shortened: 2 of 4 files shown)


hito4t:
Thank you!
Would you update README?

hiroyuki-sato:
OK. Thanks. I'll update it later.

hiroyuki-sato:
@hito4t I updated README. Please take a look when you can.

hito4t:
Thank you!
I'm sorry, I had forgotten the following question.
What is the reason for upgrading MySQL connector?

hiroyuki-sato:
I'll double check later. 
As far as I remember It's #106 related issue.

I cannot connect mysql server without `serverTimezone` using Connector/J 5.1.34.
Do I need to create new PR for update Connect/J?

|---------|-----|--------------------------------------|
| 5.1.34    | NG  | `{ useLegacyDatetimeCode: false }`   |
| 5.1.34    | OK | `{ useLegacyDatetimeCode: false, serverTimezone: UTC }`   |
| 5.1.42    | OK  | `{ useLegacyDatetimeCode: false }`   |

[My Tweet](https://twitter.com/hiroysato/status/867330496962805760)

hito4t:
@hiroyuki-sato 
Thank you!
The results may be caused by the following change in Connector/J 5.1.35 .
> A java.sql.date value was stored incorrectly on the server and also returned incorrectly if the client and the server were in different time zones when useLegacyDatetimeCode=false or useTimezone=true. This was due to the time-zone conversion performed by Connector/J on the SQL DATE type. To avoid the issue, a new property noTimezoneConversionForDateType has been created for Connector/J, which is set to “true” by default, preventing Connector/J to perform the kind of time-zone conversion that caused this bug. 

https://dev.mysql.com/doc/relnotes/connector-j/5.1/en/news-5-1-35.html

We should change warning message if updating Connector/J .
> You may need to set options `useLegacyDatetimeCode` and `serverTimezone`

https://github.com/embulk/embulk-input-jdbc/blob/master/embulk-input-mysql/src/main/java/org/embulk/input/MySQLTimeZoneComparison.java#L60

hiroyuki-sato:
@hito4t Thank you for your comment. 

I'll read `noTimezoneConversionForDateType` usage. 
Let's plan for feature implementation about embulk-ouput-mysql.

What do you think the following plan?

- First update
  - Introduce comparison timezone between server and client. 
  - If both timezones are different, log message `useLegacyDatetimeCode and serverTimezone`
  - Add warning message like `next version will update Connector/J. please take care..`
  - Introduce SSL option. 
  - Don't update Connector/J
- Second update (a few week or month later after the first update)
  - Update Connector/J  5.1.42 or later
  - Change the log message `useLegacyDatetimeCode and serverTimezone`

hito4t:
@hiroyuki-sato 
Thank you for your suggestion!
I agree.

By the way,
> Introduce comparison timezone between server and client.
> If both timezones are different, log message useLegacyDatetimeCode and serverTimezone
they were already released in 0.8.3 .

hiroyuki-sato:
@hito4t 
I reverted Connector/J version to 5.1.4.
Please take a look again. 
I'll separate new PR for output Connector/J version warning. 

I've missed seeing 0.8.3 release.

hiroyuki-sato:
Thank you!

PR #110, opened 2017-06-10
[WIP] Output a version of MySQL Connector/J.

hiroyuki-sato:
### Overview

I would like to update MySQL Connector/J to a newer version. (5.1.42 or newer)
But It has incompatibility change. 
We(I and @hito4t) decided that the plugin output a warning message like `This plugin will update Connector/J on next release...`
This PR for implement this warning. 

## Detail

@hito4t, Could you give me some advice?

- When output a version information and warning?
- How to output current Connector/J version.
  - How to get current Driver version (hard code, get version info from libraries)
- How to implement it. 

### When output a version information and warning?

I think I should the message should output before `Loaded plugin embulk/input/mysql from a load path`.

Example. 

```
2017-06-10 11:04:02.020 +0900: Embulk v0.8.22
2017-06-10 11:04:14.232 +0900 [INFO] (0001:preview): MySQL Connector/J version 5.1.34
2017-06-10 11:04:13.835 +0900 [INFO] (0001:preview): Loaded plugin embulk/input/mysql from a load path
.... and some warning message here
```

### How to output current Connector/J version?

I would like to output warning message with **current MySQL Connector/J version**.
So I tried to get it. 

The current experimental implementation get version info from `Package` object. 

First I tried to get the version info from `Driver` class. 
But It can't get sub minor version (z part in x.y.z). 
That's why I used this way. 

Do you have any alternative idea?

```java
Driver driver;
try {
    driver = new com.mysql.jdbc.Driver();

    int major = driver.getMajorVersion();
    int minor = driver.getMinorVersion();
    logger.info(String.format(Locale.ENGLISH,"driver version %d.%d",major,minor));
}
catch (SQLException e) {
    throw Throwables.propagate(e);
}
```

### DIFF
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -185,6 +185,7 @@ public abstract class AbstractJdbcInputPlugin
         Schema schema;
         try (JdbcInputConnection con = newConnection(task)) {
             // TODO incremental_columns is not set => get primary key
+            con.showDriverVersion();
             schema = setupTask(con, task);
         } catch (SQLException ex) {
             throw Throwables.propagate(ex);
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -184,8 +184,9 @@ public abstract class AbstractJdbcInputPlugin
 
         Schema schema;
         try (JdbcInputConnection con = newConnection(task)) {
-            // TODO incremental_columns is not set => get primary key
             con.showDriverVersion();
+
+            // TODO incremental_columns is not set => get primary key
             schema = setupTask(con, task);
         } catch (SQLException ex) {
             throw Throwables.propagate(ex);
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
@@ -388,7 +388,7 @@ public class JdbcInputConnection
         }
     }
 
-    void showDriverVersion(){
+    public void showDriverVersion(){
         try {
             DatabaseMetaData meta = connection.getMetaData();
             logger.info(String.format(Locale.ENGLISH,"Using JDBC Driver %s",meta.getDriverVersion()));
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
@@ -7,6 +7,7 @@ import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.Locale;
 import java.util.Set;
 
 import org.embulk.config.ConfigException;
@@ -384,6 +385,15 @@ public class JdbcInputConnection
                 columnNamesBuilder.add(rs.getString("COLUMN_NAME"));
             }
             return columnNamesBuilder.build();
+        }
+    }
+
+    void showDriverVersion(){
+        try {
+            DatabaseMetaData meta = connection.getMetaData();
+            logger.info(String.format(Locale.ENGLISH,"Using JDBC Driver %s",meta.getDriverVersion()));
+        } catch( SQLException e ) {
+            logger.warn(String.format(Locale.ENGLISH,"Can't get JDBC Driver version. Reason: %s",e.toString()));
         }
     }
 }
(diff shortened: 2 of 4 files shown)


hito4t:
Thank you for the PR!

> How to output current Connector/J version.
> How to get current Driver version (hard code, get version info from libraries)

How to implement it.

We can get driver information by `java.sql.DatabaseMetaData` object gotten by `java.sql.Connection#getMetaData()` .

`DatabaseMetaData` of MySQL Connector/J will return following values.
- `getDriverMajorVersion()` -> 5
- `getDriverMinorVersion()` -> 1
- `getDriverName()` -> "MySQL Connector Java"
- `getDriverVersion()` -> "mysql-connector-java-5.1.34 ( Revision: jess.balint@oracle.com-20141014163213-wqbwpf1ok2kvo1om )"

We should output a result of `getDriverVersion()` because other methods won't return enough information.

> When output a version information and warning?

I think we should do in `setupTask` method before checking timezone.

And it must be useful if driver version is shown in not also embulk-output-mysql but also all embulk-output-jdbc plugins.
For example, we can do in `AbstractJdbcInputPlugin#transaction` method before calling `setupTask` .

hiroyuki-sato:
@hito4t Thank you for your comment. 

I used `java.sql.Connection#getMetaData()`
What do you think this implementation?

I'll change later the following part
`This plugin will update MySQL Connector/J version on next release...``

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
```diff
@@ -386,4 +387,13 @@ public boolean tableExists(String tableName) throws SQLException
             return columnNamesBuilder.build();
         }
     }
+
+    public void showDriverVersion(){
+        try {
+            DatabaseMetaData meta = connection.getMetaData();
+            logger.info(String.format(Locale.ENGLISH,"Using JDBC Driver %s",meta.getDriverVersion()));
+        } catch( SQLException e ) {
```
hito4t: I think we don't need to catch SQLException because Connection#getMetaData#getDriverVersion doesn't throw an exception normally.
If thrown, we can't access database after all.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -185,6 +185,7 @@ public ConfigDiff transaction(ConfigSource config,
         Schema schema;
         try (JdbcInputConnection con = newConnection(task)) {
             // TODO incremental_columns is not set => get primary key
+            con.showDriverVersion();
```
hito4t: The comment above (`// TODO incremental_columns ...`) should be above the following line (`schema = setupTask(con, task);`).

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
hiroyuki-sato: Throw the exception instead of `catch`.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
hiroyuki-sato: Move comment to proper place.

hiroyuki-sato:
@hito4t Thank you for reviewing.

Please take a look again. 
I fixed your review and add log detail.

It output the message like the below. 

```
2017-06-14 14:11:22.435 +0900 [WARN] (0001:preview): This plugin will update MySQL Connector/J version on a future release.
2017-06-14 14:11:22.435 +0900 [WARN] (0001:preview): It has some incompatibility change.
2017-06-14 14:11:22.435 +0900 [WARN] (0001:preview): Please read a document and make sure configuration carefully before update the plugin.
```

hito4t:
Thank you!
Shall we elaborate the waning message?
- All plugins must upgrade drivers one day. I think "near future" is better than "future".
- It would be useful if users could know what "incompatibility" is (as written in https://github.com/embulk/embulk-input-jdbc/pull/109#issuecomment-306086994).
- I think "Please read a document ..." is not needed because users should read a document even if driver is not upgraded.

hiroyuki-sato:
@hito4t 
Thank you!
What do you think this message?
I read release note again. It seems that the incompatibility change are `noTimezoneConversionForDateType` and `cacheDefaultTimezone ` only. 

> This plugin will update MySQL Connector/J version in the near future release.
> It has some incompatibility change.
> For example, the 5.1.35 introduce `noTimezoneConversionForDateType` and `cacheDefaultTimezone ` options. 
> Please confirm your configuration before update the plugin.

hito4t:
@hiroyuki-sato 
Thank you!
That's nice.
Please commit the source after correcting errors.
some incompatibility change -> some incompatibility changes
the 5.1.35 introduce -> the 5.1.35 introduced
before update -> before updating

hiroyuki-sato:
@hito4t 

Thank you for your comment!.

I fixed some errors.
Please take a look again when you can.

hiroyuki-sato:
Thank you!

hito4t:
I'd like to resolve #113 before upgrading MySQL Connector/J.

PR #116, opened 2017-06-30
Enabled to change MySQL JDBC driver 

hito4t:
Users can use MySQL JDBC driver other than bundled JDBC driver.
We will upgrade bundled JDBC driver, but users could use old JDBC driver for compatibility.
Or users could use newer JDBC driver before bundled JDBC driver is upgraded.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -69,6 +69,7 @@ subprojects {
 
     configurations {
         provided
+        driver
     }
 
     sourceCompatibility = 1.7
@@ -110,10 +111,18 @@ subprojects {
 
     task classpath(type: Copy, dependsOn: ["jar"]) {
         doFirst { file('classpath').deleteDir() }
-        from (configurations.runtime - configurations.provided + files(jar.archivePath))
+        from (configurations.runtime - configurations.provided - configurations.driver + files(jar.archivePath))
         into 'classpath'
     }
-    clean { delete 'classpath' }
+    task driver(type: Copy, dependsOn: ["jar"]) {
+        doFirst { file('lib/embulk/driver').deleteDir() }
+        from (configurations.driver)
+        into 'lib/embulk/driver'
+    }
+    clean {
+        delete 'classpath'
+        delete 'lib/embulk/driver'
+    }
 
     checkstyle {
         configFile = file("${project.rootDir}/config/checkstyle/checkstyle.xml")
@@ -132,7 +141,7 @@ subprojects {
         source = sourceSets.main.allJava + sourceSets.test.allJava
     }
 
-    task gem(type: JRubyExec, dependsOn: ['build', 'gemspec', 'classpath']) {
+    task gem(type: JRubyExec, dependsOn: ['build', 'gemspec', 'classpath', 'driver']) {
         jrubyArgs '-rrubygems/gem_runner', "-eGem::GemRunner.new.run(ARGV)", 'build'
         scriptArgs "${project.projectDir.absolutePath}/build/gemspec"
         doLast { ant.move(file: "${project.name}-${project.version}.gem", todir: "${parent.projectDir}/pkg") }
@@ -186,7 +195,7 @@ Gem::Specification.new do |spec|
   spec.licenses      = ["Apache 2.0"]
   spec.homepage      = "https://github.com/embulk/embulk-input-jdbc"
 
-  spec.files         = `git ls-files`.split("\n") + Dir["classpath/*.jar"]
+  spec.files         = `git ls-files`.split("\n") + Dir["classpath/*.jar"] + Dir["lib/embulk/driver/*.jar"]
   spec.test_files    = spec.files.grep(%r"^(test|spec)/")
   spec.require_paths = ["lib"]
 end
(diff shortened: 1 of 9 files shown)


hiroyuki-sato:
It worked in my environment.

loaded `Using JDBC Driver mysql-connector-java-5.1.42`.


```
embulk preview -I /path/to/embulk-input-jdbc/embulk-input-mysql/lib config2.yml.liquid
2017-07-03 12:55:52.621 +0900: Embulk v0.8.25
2017-07-03 12:55:54.085 +0900 [INFO] (0001:preview): Loaded plugin embulk/input/mysql from a load path
2017-07-03 12:55:54.119 +0900 [INFO] (0001:preview): Fetch size is 10000. Using server-side prepared statement.
2017-07-03 12:55:54.298 +0900 [INFO] (0001:preview): Using JDBC Driver mysql-connector-java-5.1.42 ( Revision: 1f61b0b0270d9844b006572ba4e77f19c0f230d4 )
2017-07-03 12:55:54.298 +0900 [WARN] (0001:preview): This plugin will update MySQL Connector/J version in the near future release.
2017-07-03 12:55:54.298 +0900 [WARN] (0001:preview): It has some incompatibility changes.
2017-07-03 12:55:54.298 +0900 [WARN] (0001:preview): For example, the 5.1.35 introduced `noTimezoneConversionForDateType` and `cacheDefaultTimezone` options.
2017-07-03 12:55:54.298 +0900 [WARN] (0001:preview): Please read a document and make sure configuration carefully before updating the plugin.
2017-07-03 12:55:54.300 +0900 [WARN] (0001:preview): The plugin will set `useLegacyDatetimeCode=false` by default in future.
2017-07-03 12:55:54.403 +0900 [INFO] (0001:preview): Fetch size is 10000. Using server-side prepared statement.
2017-07-03 12:55:54.408 +0900 [INFO] (0001:preview): SQL: SELECT * FROM `time_test`
2017-07-03 12:55:54.411 +0900 [INFO] (0001:preview): > 0.00 seconds
+---------+-------------------------+-------------------------+-------------------------+
| id:long |   c_date_time:timestamp |        c_date:timestamp |   c_timestamp:timestamp |
+---------+-------------------------+-------------------------+-------------------------+
|       1 | 2017-06-30 15:00:00 UTC | 2017-06-30 15:00:00 UTC | 2017-06-30 15:00:00 UTC |
+---------+-------------------------+-------------------------+-------------------------+
```


```yaml
in:
  type: mysql
  table: time_test
  database: embulk_test
  user: root
  password: {{ env.mysql_password }}
  host: 127.0.0.1
  driver_path: /tmp/mysql-connector-java-5.1.42.jar

out:
  type: stdout
```

muga:
@hito4t This feature sounds good for me:-) 

Do you have any reasons why you decided to put a bundled driver on `lib/embulk/driver`? The `lib/embulk/driver` works fine for `JRubyPluginSource` but, the path might not work for `MavenPluginSource`, which is new mechanism to load plugins. It enables loading jar files as plugins. To load the plugin by both of them, I think that it's better to put the bundled driver's jar file to `src/main/resources/xxx`. What do you think about it?  cc: @dmikurube

dmikurube:
@muga `MavenPluginSource` does not load such a JAR-in-JAR anyway unless we prepare a special `ClassLoader` like One-JAR: http://one-jar.sourceforge.net/ The special `ClassLoader` would be in the plugin, not in the core. And then, JARs-in-JAR can be loaded from anywhere in the main plugin JAR.

Anyways, to build a JAR for `MavenPluginSource`, we have to have a special `build.gradle` and a special `ClassLoader`, which are totally different from this `.gem` version. I don't think we need to take care of the `.jar` version at the moment.

@hito4t Then, just for this PR for the `.gem` version, I thought `lib/embulk/driver` is not very intuitive because we often think:
1) Java plugins have only Ruby code under `/lib/embulk`.
2) JAR files are usually under `/classpath` just under the root.
3) Not very sure what does the directory name `driver` mean.

IMHO, locating such driver JAR file(s) in a directory just under the root, such as below:
* `/default_jdbc/mysql-connector-java-5.1.42.jar`
* `/default_jdbc_driver/mysql-connector-java-5.1.42.jar`

What do you think?

The other parts look good to me.

hito4t:
@muga @dmikurube 
Thank you for your comments!

> Do you have any reasons why you decided to put a bundled driver on lib/embulk/driver? 

Because I put native libraries for Oracle on `lib/embulk/native`, but it may not be appropriate for JDBC drivers.

> IMHO, locating such driver JAR file(s) in a directory just under the root, such as below:
>    /default_jdbc/mysql-connector-java-5.1.42.jar
>    /default_jdbc_driver/mysql-connector-java-5.1.42.jar

I agree.
I think the latter is clearer.

hito4t:
Now developer should download and put some JDBC drivers on `driver` folder for Gradle test because embulk cannot bundle them for reasons of license.
The location of JDBC driver is written in `build.gradle`.

I've tried to specify it by `driver_path` property in yml so that developers could locate it anywhere, but couldn't do because ClassLoader for Gradle test is default Java ClassLoader, not PluginClassLoader.

Anyway, I'll change `driver` to `test_jdbc_driver`.

muga:
LGTM 👍

PR #119, opened 2017-09-29
Fix column name in example query

msysyamamoto:
Change `created_at` to `updated_at`.


hito4t:
Thank you!

hito4t:
@muga 
Can you check why CI failed?

PR #120, opened 2017-11-13
Support array column type

instcode:
This PR implements `array` column type support for Postgres.

The implementation supports converting both single/multiple dimension array to string & json. However, in json case, only simple value types (number, bool, and text) are allowed in the array elements, e.g. integer[], text[][], bigint[][][]...

The implementation does have a compatible limitation. It works perfectly for Postgres server version 8.3.0 and above due to the usage of object type versus primitive type for array elements. This can be fixed but it would introduce a very complex code base.

### DIFF
--- a/embulk-input-postgresql/README.md
+++ b/embulk-input-postgresql/README.md
@@ -35,7 +35,7 @@ PostgreSQL input plugin for Embulk loads records from PostgreSQL.
 - **default_timezone**: If the sql type of a column is `date`/`time`/`datetime` and the embulk type is `string`, column values are formatted int this default_timezone. You can overwrite timezone for each columns using column_options option. (string, default: `UTC`)
 - **column_options**: advanced: a key-value pairs where key is a column name and value is options for the column.
   - **value_type**: embulk get values from database as this value_type. Typically, the value_type determines `getXXX` method of `java.sql.PreparedStatement`.
-  (string, default: depends on the sql type of the column. Available values options are: `long`, `double`, `float`, `decimal`, `boolean`, `string`, `json`, `date`, `time`, `timestamp`)
+  (string, default: depends on the sql type of the column. Available values options are: `long`, `double`, `float`, `decimal`, `boolean`, `string`, `json`, `date`, `time`, `timestamp`, `array`)
   See below for `hstore` column.
   - **type**: Column values are converted to this embulk type.
   Available values options are: `boolean`, `long`, `double`, `string`, `json`, `timestamp`).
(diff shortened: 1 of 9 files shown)


instcode:
Hi @hito4t, could you please have a look? Thank you.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/ArrayColumnGetter.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        catch (JsonParseException | SQLException | ClassCastException e) {
+            super.jsonColumn(column);
+        }
+    }
+
+    @Override
+    public void stringColumn(Column column)
+    {
+        to.setString(column, value.toString());
```
hito4t: java.sql.Array is interface and doesn't implement toString method.
We should implement the code converting Array to String by hand, otherwise the result will vary depends on DBMS.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/ColumnGetterFactory.java
```diff
@@ -62,6 +62,8 @@ private ColumnGetter newColumnGetter(JdbcInputConnection con, PluginTask task, J
             return new TimestampColumnGetter(to, toType, newTimestampFormatter(option, DateColumnGetter.DEFAULT_FORMAT));
         case "decimal":
             return new BigDecimalColumnGetter(to, toType);
+        case "array":
```
hito4t: Because "array" type is specific for PostgreSQL, these lines should be in PostgreSQLColumnGetterFactory class.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/ArrayColumnGetter.java
instcode: Yes, because the implementation is kinda specific for Postgres, I refactored the code from `embulk-input-jdbc` to `embulk-input-postgresql`.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/ColumnGetterFactory.java
instcode: Got it. I moved to `embulk-input-postgresql`.

instcode:
Thanks @hito4t for your review comments. I refactored the code and added unit-test to handle the `array` type specific for Postgres. I also updated the README to add more information about this support. Please have a look ^^ Thanks.

instcode:
BTW, the tests run okay on my machine but it seems Travis says no :-). The build stops at `embulk-input-mysql` errors though.

sakama:
@instcode I think Travis CI failure will be fixed by https://github.com/embulk/embulk-input-jdbc/pull/122

hito4t:
@instcode
Great!
Thank you!

PR #128, opened 2017-12-20
fix "datetime" type timezone convert support

jo8937:
sqlserver's datetime field cannot convert timezone Asia/Seoul to UTC


jo8937:
oh, I'm Sorry. It has bug for use default format is equal to date type.

PR #135, opened 2018-01-31
Fix null value when building postgresql array column value to string

puppycloud:
There are 2 ways to create an instance of `org.postgresql.jdbc4.Jdbc4Array` when reading from a ResultSet. Here is an excerpt from `org.postgresql.jdbc4.Jdbc4ResultSet`:

    // 1st
    protected Array makeArray(int oid, byte[] value) throws SQLException {
        return new Jdbc4Array(this.connection, oid, value);
    }
    // 2nd
    protected Array makeArray(int oid, String value) throws SQLException {
        return new Jdbc4Array(this.connection, oid, value);
    }

However, in the current implementation support for Array to String output, the value of array column is read like this:

    @Override
    public void stringColumn(Column column)
    {
        to.setString(column, jdbc4Array.toString());
    }

Unfortunately, `Jdbc4Array#toString` simply returns the String `value` passed to the constructor in the 2nd "makeArray" method above and may return null in case 1st "makeArray" is used.

In my real test against local Postgres DB with multiple pages fetching, only the first 5 pages, the driver uses the 2nd way to make the Jdbc4Array instances and from the 6th page, it uses the 1st way to create the array and causes the input plugin failed to get the representation value of array column.

This PR addresses the issue above by manually generating the string value for such array column value instead of relying on the `Jdbc4Array#toString` follow the specification mentioned here: https://www.postgresql.org/docs/current/static/arrays.html


embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/getter/ArrayColumnGetter.java
```diff
@@ -72,6 +72,8 @@ private ArrayNode buildJsonArray(Object[] elements)
                     case Types.LONGNVARCHAR:
                         arrayNode.add(jsonNodeFactory.textNode((String) v));
                         break;
+                    default:
```
hito4t: `ArrayColumnGetter` didn't support array of `decimal` type.
This change supports `decimal` array, but decimal values will be converted into string values such as `"[""1.0"",""2.0"",""3.0""]"`.
Could you fix the problem?

puppycloud: I could remove this but wouldn't it be better if we use this fallback for whatever type we don't support? I mean, in the case it's a decimal or a numeric type, a string could be a perfect selection to represent its value (because it could be up to 131072 digits before the decimal point; up to 16383 digits after the decimal point: https://www.postgresql.org/docs/current/static/datatype-numeric.html). What do you think?

hito4t: @puppycloud 
I think `BigDecimal` is preferable to `String` for `decimal` type.
How about the following codes?
<pre>
    case Types.DECIMAL:
    case Types.NUMERIC:
        arrayNode.add(jsonNodeFactory.numberNode((BigDecimal)v));
        break;
    default:
        arrayNode.add(jsonNodeFactory.textNode(v.toString()));
        break;
</pre>

puppycloud: Yes, that's absolutely better. I updated my unit-test to cover this as well. Thanks.

hito4t:
@puppycloud 
Thank you!

PR #136, opened 2018-03-28
Support UUID postgresql

trung-huynh:
UUID will now be acted as string instead of having the specific column_option for that


hito4t:
Thank you for the PR!
It must be useful.

PR #137, opened 2018-05-22
Add table hint option to sqlserver plugin.

kmrshntr:
```
in:
・・・
  table: TABLE_NAME
  table_hint: NOLOCK
  where: "updated_at <= DATEADD(hour, -1, getdate())"
・・・
```

The above configuration generates following SQL:

```
SELECT * FROM "TABLE_NAME" with (NOLOCK) WHERE updated_at <= DATEADD(hour, -1, getdate())
```

### DIFF
--- a/embulk-input-sqlserver/README.md
+++ b/embulk-input-sqlserver/README.md
@@ -27,6 +27,7 @@ embulk "-J-Djava.library.path=C:\drivers" run input-sqlserver.yml
   - **query**: SQL to run (string)
 - If **query** is not set,
   - **table**: destination table name (string, required)
+  - **table_hint**: table hint (e.g. `NOLOCK`) (string, optional)
   - **select**: expression of select (e.g. `id, created_at`) (string, default: "*")
   - **where**: WHERE condition to filter the rows (string, default: no-condition)
   - **order_by**: expression of ORDER BY to sort rows (e.g. `created_at DESC, id ASC`) (string, default: not sorted)
--- a/embulk-input-sqlserver/README.md
+++ b/embulk-input-sqlserver/README.md
@@ -27,7 +27,7 @@ embulk "-J-Djava.library.path=C:\drivers" run input-sqlserver.yml
   - **query**: SQL to run (string)
 - If **query** is not set,
   - **table**: destination table name (string, required)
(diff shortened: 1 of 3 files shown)


hito4t:
@kmrshntr 
Thank you for the PR!

This feature should be useful.
It will be more useful if other DBMSs support the feature.

For example,
- add `read_uncommitted` property (false in default)
- if true, embulk-input-sqlserver will add 'WITH (NOLOCK)`
- other embulk-input-xxx will support the feature in future

What do you think about that?

kmrshntr:
@hito4t I see your point. That's better for consistency over every DBMSs.

I thought another idea.

 * Add `transaction_isolation_level` property.
 * If the value is `NOLOCK`, embulk-input-sqlserver will add ` WITH (NOLOCK)`
 * If the value is `READUNCOMMITTED`, embulk-input-sqlserver will add ` WITH (READUNCOMMITTED)`
 * If the value is `READPAST`, embulk-input-sqlserver will add ` WITH (READPAST)`

That is just renaming of `table_hint` property in my first idea. `transaction_isolation_level` is more natural for other DBMSs.

The embulk-input-sqlserver plugin is just for SQLServer integration, right?
So I think those flexible configurations specialized for SQLServer is also reasonable.

How about that?

hito4t:
@kmrshntr 
Thank you for your idea!
`transaction_isolation_level` is more flexible than `read_uncommitted`.

It's preferable that all embulk-input-jdbc plugins have same properties and support same property values.
(Some users use multiple plugins, and I am maintaining all plugins.)
But that is not easy because DBMSs have different transaction mechanisms.

I agree with your idea.

kmrshntr:
@hito4t I renamed the property to `transaction_isolation_level`.

Thank you for your acceptance. 🙂

hito4t:
@kmrshntr 
Thank you!

kmrshntr:
@hito4t Do you know when the new version of the gem including changes of #137 and #139 is going to be released ?

hito4t:
@kmrshntr 
I'll release the new version in a few weeks.

hito4t:
@kmrshntr 
I've released the new version.
I'm sorry for the delay.

kmrshntr:
@hito4t Perfect timing💯  Thank you!

PR #138, opened 2018-05-23
Support unknown types as string

trung-huynh:
Currently once `embulk-input-postgresql` imports unknown types from PostgreSQL e.g. `bytea`, the exception will be thrown because plugin does not know what should it does with that type.

The changes in this PR will automatically "tell" plugin consider all of unknown types as String instead of throw out the exception. I know there is one way of that by using `column_options` but by using this PR the `value_type` of string will be by default to all columns.


trung-huynh:
@hito4t Thanks for you explanation, that makes sense to use those default fields. I'll close this PR then.

PR #139, opened 2018-05-23
Support datetime type column of SQLServer for incremental_columns option.

kmrshntr:
This is just for embulk-input-sqlserver.

If `incremental_columns` option is filled with datetime type column, the exception with following message comes out in v0.9.1.

> Column type 'timestamp' set at `incremental_columns` option is not supported

The columns whose type is datetime is supported as `incremental_columns` option by this PR.


hito4t:
Thank you for the PR!

Issue #100, opened 2017-03-22
Allowing incremental_columns when using query option


kakoni:
Currently 'incremental_columns' and 'last_record' parameters are not supported if 'query' parameter is set.

Is there specific reason for this?
Tested few queries(with lines 308-309 removed from AbstractJdbcInputPlugin.java) and things worked as expected.

hito4t:
Hi
First, the plugin built query for incremental loading from table name and didn't support `query` parameter.
After that, the plugin came to be able to build query for incremental loading from any query, but the check was not deleted.

I think we can delete the check.
https://github.com/embulk/embulk-input-jdbc/blob/master/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java#L326-L327

@frsyuki
What do you think about it?

frsyuki:
With `incremental_columns` option, the plugin needs to generate add a WHERE condition and ORDER BY clause. When `query` option is given, `JdbcInputConnection#wrapIncrementalQuery` method runs and it adds WHERE and ORDER BY by wrapping the original `query` in a subquery. So, `query` option with `incremental_columns` is an expected use case. I think it's ok to remove the outdated validation, as @hito4t mentions.

A concern is that wrapping a query in a subquery spoils utilization of table indexes in MySQL (in some cases) although the generated query seems totally reasonable to use an index on ORDER BY column(s). This concerns me because users won't expect that using `query` option and not using `query` option won't make significant differences (but actually they do). And they'll realize this problem only when Embulk made the database overloaded, in current implementation.

It would be great if there is an idea to mitigate this unexpected behavior, such as adding a document or warning message. For PostgreSQL, Redshift, or other databases that have reasonable query optimizer, I don't have concerns.

y-ken:
Hi @frsyuki @hito4t 
I need this enhancement so much.
To avoid unexpected behavior, I suggest to introduce a option or placeholder.
How about that?

* If it does not have placeholder, just keep current subquery behavior
* If it has a option to disable subquery
  * If the query already has `ORDER BY`, warn it end exit.
  * If the query contain `WHERE`, wrap current condition with `()`and add `AND id > ?`
  * Add `ORDER BY` statement.

For example
```
-- input
SELECT
  foo.id, bar.name
FROM
  foo LEFT JOIN bar ON foo.id = bar.id
WHERE
  hoge IS NOT NULL

-- output
SELECT
  foo.id, bar.name
[...]

PR #143, opened 2018-06-05
Allow incremental load with 'query' option

sakama:
This PR will solve #100 

## How does this work
I added new option `use_raw_query(boolean, default:false)`.
This feature works only when `use_raw_query: true` and `incremental: true`.

### Example of config.yml
```yaml
in:
  type: mysql
  host: 127.0.0.1
  user: {{ env.mysql_user }}
  password: {{ env.mysql_password }}
  query: |    
    SELECT
      foo.id as foo_id, bar.name
    FROM
      foo LEFT JOIN bar ON foo.id = bar.id
    WHERE
      foo.hoge IS NOT NULL
      AND foo.id > :foo_id
    ORDER BY
      foo.id ASC
  incremental_columns:
  - foo_id
  incremental: true
  use_raw_query: true
  last_record: [1]

out:
  type: stdout
```

**Plugin will replace `:foo_id` with last_record's value '1'.**
**Plugin will also generate new last_record after execution.** This behavior is same with current incremental loading.

## Restrictions
* Plugin user have to understand behavior of incremental load of Embulk and embuk-input-jdbc/*

All of below settings are needed when `use_raw_query:true`
* `query` option is required (validation exits)
* `last_record` is required (validation exits)
* `last_record` size should be same with size of `incremental_columns` (validation exits)

## Remaining tasks
* Test
* Update unit test
* Update README (I'm wondering if this feature should be officially announced or not).


sakama:
I'm updating unit test for now.
Any comments from everyone for specifications and implementations are welcome.

sakama:
Added unit test for PostgreSQL, MySQL, RedShift, Oracle.
@hito4t Can I ask you your opinion?

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -67,6 +67,10 @@
         @ConfigDefault("null")
         public Optional<String> getQuery();
 
+        @Config("use_raw_query")
+        @ConfigDefault("false")
+        public boolean getUseRawQuery();
```
hito4t: This property will be used only for incremental loading.
Users will understand easily if the property name shows that (for example, it contains `incremental`).

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
```diff
@@ -351,6 +356,22 @@ public PreparedQuery wrapIncrementalQuery(String rawQuery, JdbcSchema querySchem
...
 
+    private List<JdbcLiteral> replacePlaceholder(StringBuilder sb, String rawQuery, JdbcSchema querySchema,
+                                                 List<Integer> incrementalColumnIndexes, List<JsonNode> incrementalValues)
+    {
+        ImmutableList.Builder<JdbcLiteral> parameters = ImmutableList.builder();
+        for (int n = 0; n < incrementalColumnIndexes.size(); n++) {
+            int columnIndex = incrementalColumnIndexes.get(n);
+            String columnName = querySchema.getColumnName(columnIndex);
+            JsonNode value = incrementalValues.get(n);
+            rawQuery = rawQuery.replaceAll(":" + columnName, "?");
```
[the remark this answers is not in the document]
sakama: Sorry to be late.

> In some cases it won't work correctly.

I think using something SQL parser is too much in this case.
What do you think about replace with regular expression?

```java
rawQuery = rawQuery.replaceAll(":" + columnName + "[^\'\"]+?", " ? ");
```
At least, "replace longer column names previously" problem will be solved.
The pattern of regex is arguable and isn't a perfect solution, but.

> At least we should add annotation about it to README.

OK. I'll add it.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
sakama: How about `use_query_with_incremental`?
Or do you have any favorite candidates?

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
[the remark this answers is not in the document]
hito4t: > I think using something SQL parser is too much in this case.

I think so too.

> What do you think about replace with regular expression?

Replacing with regular expression is difficult because 
- DBMSs have various special characters. 
- column name might contain special characters (ex. `$`).

And I found that the method doesn't work correctly in the following cases.
- Same column name appears multiple times. Parameters will be short.
`... where :a > x or :a = x and :b > y`

- The order of column names in incremental_columns and SQL are different.
<pre>
incremental_columns : [a, b]
query: ... where :b > y and :a > x</pre>

How about the following alogrithm?

1. create map whose keys are the column names and whose values are JdbcLiteral objets.
2. sort the column names in the order of longer lengths.
3. find ':' from the beginning of the query string.
4. if the rest of the query string begins with one of the column names, replace with '?' and add JdbcLiteral object to parameters.
5. find ':' from the rest of the query string. -> 4

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
hito4t: Thank you for the proposition.
How about `use_raw_query_with_incremental` ?
`query` will be used even if the property is `false`.
(But the property name is a bit longer...)

sakama: Sounds good. I'm going to replace with `use_raw_query_with_incremental`

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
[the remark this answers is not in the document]
sakama: OK. Let me implement with your suggestion.

sakama:
I updated PR based on discussion.
@hito4t What do you think about?

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -210,7 +216,29 @@ protected Schema setupTask(JdbcInputConnection con, PluginTask task) throws SQLE
...
+                    return val2.length() - val1.length();
+                }
+            });
+            for (int i = 0; i < task.getIncrementalColumns().size(); i++) {
+                columnNames.put(task.getIncrementalColumns().get(i), i);
+            }
+
+            for (Map.Entry<String, Integer> column : columnNames.entrySet()) {
+                // Temporary replace place holder like ":id" with "?" to avoid SyntaxException while getting schema.
+                temporaryQuery = temporaryQuery.replaceAll(":" + column.getKey(), "?");
```
hito4t: We should use `replace` instead of `replaceAll`.

embulk-input-jdbc/README.md
```diff
@@ -76,6 +77,38 @@ CREATE INDEX embulk_incremental_loading_index ON table (updated_at, id);
 
 Recommended usage is to leave `incremental_columns` unset and let this plugin automatically finds an auto-increment primary key. Currently, only strings and integers are supported as incremental_columns.
 
+### Use incremental loading with complex query
+
+**IMPORTANT**: This is an advanced feature and assume you have an enough knowledge about incremental loading using Embulk and this plugin
+
+Normally, you can't write your own query for incremental loading.
+`use_raw_query_with_incremental` option allow you to write complex query for incremental loading.
```
hito4t: The reason to use `use_raw_query_with_incremental` is writing optimized (faster) query, isn't it?
https://github.com/embulk/embulk-input-jdbc/issues/100#issuecomment-289352025

embulk-input-mysql/src/test/java/org/embulk/input/mysql/IncrementalTest.java
```diff
@@ -104,4 +104,24 @@ public void testTimestamp() throws Exception
         assertThat(readSortedFile(out2), is(readResource("ts/expected_2.csv")));
         assertThat(result2.getConfigDiff(), is((ConfigDiff) loadYamlResource(embulk, "ts/expected_2.diff")));
     }
+
+    @Test
+    public void testQueryWithPlaceholder() throws Exception
```
hito4t: Can you add a test for multiple incremental columns?

sakama: Fixed at 9fe5f87b02e26b2f3b27b0cae7238b8d38e62c58 and added for embulk-input-(oracle|db2|sqlserver|postgresql|oracle|redshift) dc0f798c668eb44e2ced2075ddf5c882e3b48d3a

embulk-input-jdbc/README.md
sakama: Fixed at 9fe5f87b02e26b2f3b27b0cae7238b8d38e62c58 and added for embulk-input-(oracle|db2|sqlserver|postgresql|oracle|redshift) dc0f798c668eb44e2ced2075ddf5c882e3b48d3a

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
sakama: Fixed at dbc6e380269aae3559528aa6d82f21c18a452619

sakama:
Now, I found a bug in current logic which generates parameter(JDBCLiteral).
This problem happens when config has multiple columns.
Parameter will be added by unintended order -  column length order DESC.

In below example, "num2" value will be added at head of parameters because column name length of "num2" is longer than "num"
```yaml
last_record: [0,101]
query:
  SELECT
    *
  FROM
    query_load
  WHERE
    num IS NOT NULL
    AND num > :num
    OR (num = :num AND num2 > :num2)
  ORDER BY
    num ASC
```
expected parameters: [0,0,101]
actual parameter values: [101,0,0]

I'm considering what is the best solution for it.

sakama:
I updated parameters addition logic 134385fe1e361d221559fe686b00032fef099c17

First, plugin will parse raw query using `String.indexOf(":num2", lastIndex)` method and put `index position : columnIndex` pair to Map.
This value will be sorted by index position ASC.
e.g. [65,0][105,0][121,1]

```java
// variable columnNames has Map value sorted by column name length DESC [num2, 1][num, 0]
for (Entry<String, Integer> column : columnNames.entrySet()) {
    int lastIndex = 0;
    while (true) {
        int index = rawQuery.indexOf(":" + column.getKey(), lastIndex);
        if (index == -1) {
            break;
        }
        if (!columnPositionList.containsKey(index)) {
            columnPositionList.put(index, column.getValue());
        }
        lastIndex = index + 2;
    }
}
```
After that, plugin will extract columnIndex from above Map value and add its value to parameters.

@hito4t What do you think about?

embulk-input-jdbc/README.md
```diff
@@ -22,6 +22,7 @@ Generic JDBC input plugin for Embulk loads records from a database using a JDBC
 - **options**: extra JDBC properties (hash, default: {})
 - If you write SQL directly,
   - **query**: SQL to run (string)
+  - **use_raw_query_with_incremental**: If true, you can write optiomized query using prepared statement. See [Use incremental loading with complex query](#use-incremental-loading-with-complex-query) for more detail (boolean, default: false)
```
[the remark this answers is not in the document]
sakama: Fixed at 713c541a56695f92cf6cfe54683ec7dc0ba23e47

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
```diff
@@ -351,6 +361,85 @@ public PreparedQuery wrapIncrementalQuery(String rawQuery, JdbcSchema querySchem
...
+    */
+    private TreeMap<Integer, Integer> generateColumnPositionList(String rawQuery, TreeMap<String, Integer> columnNames)
+    {
+        TreeMap<Integer, Integer> columnPositionList = new TreeMap<>(new Comparator<Integer>() {
+            @Override
+            public int compare(Integer val1, Integer val2) {
+                return val1 - val2;
+            }
+        });
+
```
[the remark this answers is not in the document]
sakama: Ah, thanks. Removed at ef1abbe202bafdb33331577e2882878346f259d6

embulk-input-db2/README.md
```diff
@@ -78,6 +79,38 @@ CREATE INDEX embulk_incremental_loading_index ON table (updated_at, id);
 
 Recommended usage is to leave `incremental_columns` unset and let this plugin automatically finds a primary key. Currently, only strings and integers are supported as incremental_columns.
 
+### Use incremental loading with complex query
```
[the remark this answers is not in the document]
sakama: Rewrote to `raw_query` instead of `complex_query`  52bfbd845d363e0be6d20db0e0c61aee4ea735c1

hito4t:
@sakama 
Thank you!

sakama:
👍 
I appreciate you kindly took long time for review.
Thank you!

PR #148, opened 2018-09-18
Added `before_select` to Redshift config

hitul007:
- `redshift-input-plugin` has new field called `before_select`. SQL query written in field will be executed just before select `query` executed.
- Field can be used to do inserts, create temp tables or any other use case where data is not getting returned.
- In Redshift ideally temp tables should be created with `CREATE TEMP TABLE` syntax. We can specify sort key and dist keys in mentioned way.

### DIFF
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -720,6 +720,6 @@ public abstract class AbstractJdbcInputPlugin
                 maskedProps.setProperty(key, props.getProperty(key));
             }
         }
-        logger.info("Connecting2 to {} options {}", url, maskedProps);
+        logger.info("Connecting to {} options {}", url, maskedProps);
     }
 }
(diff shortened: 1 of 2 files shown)


hito4t:
Thank you for the PR!

It will be very useful if we can execute SQL before SELECT.

In the PR, SQL will be executed three times.
- before the `setupTask` method, `temp_table_def`  will be executed
- at the beggining of the `run` method, `temp_table_def`  will be executed
- at the beggining of the `run` method, `before_select`  will be executed

I think we don't need to execute SQL in the `run` method if we can execute SQL before the `setupTask` method.
What do you think?

hitul007:
@hito4t I wanted to execute,
```
CREATE TABLE college_student (
    name     varchar(100),
    roll_no   int
)
sortkey(roll_no);


INSERT INTO student
SELECT name, roll_no
FROM college_student
where collect = 'j g college';


SELECT * FROM student
LEFT JOIN student ON (student.roll_no = collect_student.roll_no);

```

Now, If I try to execute above query then I'll get `syntax error` before of `FETCH` in query. 

Now, I'm executing `temp_table_def` at time of schema check because new connection is getting created and closed once schema is identified.  New connection is also established when we are running `query` and I'm getting an error saying temp table does not exist because after schema check connection was closed and this is a reason I'm executing temp table definitions before schema making and before query execution. 

Ideal solution will be to use single connection in schema making and before query execution but, I think this will be huge change and will affect lot of other stuff.

If you can suggest some good and easy solution then I can work on it.

hito4t:
@hitul007 
Thank you for your reply!

I understood your usecase.

As you wrote, you need to execute SQL twice.

But I think `temp_table_def` is a bit confusing because users can write any SQL (it is not easy to check if it is a DDL for creating temporary table).

How about adding two properties : `before_setup` and `before_select` ?
(in the case, you need to write the same `CREATE` statement twice (in `before_setup` and in `before_select`))

hito4t:
I've implemented `before_setup` and `before_select` in the PR #159 referring to this PR.
Thank you.

PR #153, opened 2018-12-14
Support JSON type for MySQL

tvhung83:
Currently, `json` type of MySQL will be mapped to `String` type of Embulk. This PR will map it to `json` type of Embulk. It also drop support of Java 7, and upgrade dependencies to latest Embulk version.


dmikurube:
The Java 8 part looks good to me. Concerns are:

* Compatibility from previous versions of the MySQL plugin.
* Making it Java 8 means it will run only on Embulk v0.9+.

hito4t:
@dmikurube 
Thank you for review!


> Compatibility from previous versions of the MySQL plugin.

I'm concerned with that too, but I think JSON should be converted JSON as embulk-input-postgresql does.
The older embulk-input-mysql was not able to support JSON type because the bundled Connector/J (5.1.34) doesn't support MySQL 5.7, which begins to support JSON type.
But now we can support JSON type in default.
(Actually, we was able to support JSON type in embulk-input-mysql 0.8.5, which can use any JDBC driver.)

We should announce and warn about the incompatibility.

@tvhung83 
- It will work, but would you handle JSON type in `sqlTypeToValueType` method, not in `newColumnGetter` method, as `org.embulk.input.postgresql.getter.PostgreSQLColumnGetterFactory` does?

- Can you add `testJson` to `org.embulk.input.mysql.BasicTest` ?

tvhung83:
@hito4t 
Thank you for your comments, will add and confirm here again.

tvhung83:
@hito4t 
Sorry for late update. Please take a look again when you got a chance, thanks.

hito4t:
@tvhung83 
Thank you!
It seems very good!

@dmikurube 
Can you review ".travis.yml"?

hito4t:
@dmikurube 
@sakama 
Thank you!

PR #156, opened 2019-01-24
Fix example for use_raw_query_with_incremental

szyn:
Hi, I've added pipe (`|`) to query parameter.


szyn:
Oh, I have a mistake when rebasing...😢  
I'll create as another PR so this PR close for now.

PR #161, opened 2019-05-15
Fix missing left brace when getting empty array column

ommadawn46:
I fixed a problem that left braces were missing when getting an empty array with `embulk-input-postgresql`.

## Insert data to a source table

```
insert into input_array (c1, c2, c3, c4) values ('{}', '{}', '{}', '{}');
```

## Result before fix

```
},},},}
```

## Result after fix

```
{},{},{},{}
```


hito4t:
@ommadawn46 
Thank you for fixing the bug!

PR #170, opened 2019-09-02
Fix typo of Ubuntu distribution name in .travis.yml to fix CI failure

sakama:
s/xernial/xenial to fix Travis CI failure
https://travis-ci.org/embulk/embulk-input-jdbc/builds/577710116


hito4t:
@sakama 
Thank you!

PR #172, opened 2019-12-03
Update ssl option doc README

y-ken:
There are ssl option implementation but README does not have the information.
https://github.com/embulk/embulk-input-jdbc/blob/master/embulk-input-mysql/src/main/java/org/embulk/input/MySQLInputPlugin.java#L50-L52

### DIFF
--- a/embulk-input-jdbc/README.md
+++ b/embulk-input-jdbc/README.md
@@ -14,7 +14,6 @@ Generic JDBC input plugin for Embulk loads records from a database using a JDBC 
 - **url**: URL of the JDBC connection (e.g. 'jdbc:sqlite:mydb.sqlite3') (string, required)
 - **user**: database login user name (string, optional)
 - **password**: database login password (string, default: optional)
-- **ssl**: use SSL to connect to the database (string, default: `disable`. `enable` uses SSL without server-side validation and `verify` checks the certificate. For compatibility reasons, `true` behaves as `enable` and `false` behaves as `disable`.)
 - **schema**: destination schema name (string, default: use the default schema)
 - **fetch_rows**: number of rows to fetch one time (integer, default: 10000)
 - **connect_timeout**: not supported.
--- a/embulk-input-mysql/README.md
+++ b/embulk-input-mysql/README.md
@@ -39,6 +39,7 @@ MySQL input plugin for Embulk loads records from MySQL.
     - Internally, `useCursorFetch=false` is used and `java.sql.Statement.setFetchSize` is not set.
 - **connect_timeout**: timeout for socket connect. 0 means no timeout. (integer (seconds), default: 300)
 - **socket_timeout**: timeout on network socket operations. 0 means no timeout. (integer (seconds), default: 1800)
+- **ssl**: use SSL to connect to the database (string, default: `disable`. `enable` uses SSL without server-side validation and `verify` checks the certificate. For compatibility reasons, `true` behaves as `enable` and `false` behaves as `disable`.)
 - **options**: extra JDBC properties (hash, default: {})
 - **incremental**: if true, enables incremental loading. See next section for details (boolean, default: false)
 - **incremental_columns**: column names for incremental loading (array of strings, default: use primary keys). Columns of integer types, string types, `datetime` and `timestamp` are supported.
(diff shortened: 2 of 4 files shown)


y-ken:
@sakama or @hito4t  could you please check this?

hito4t:
@y-ken 
Thank you for the PR!

The `ssl` option of embulk-input-mysql is same as the `ssl` option of embulk-output-mysql.
How about copying readme from embulk-output-mysql?
https://github.com/embulk/embulk-output-jdbc/blob/master/embulk-output-mysql/README.md
<pre>
- **ssl**: use SSL to connect to the database (string, default: `disable`. `enable` uses SSL without server-side validation and `verify` checks the certificate. For compatibility reasons, `true` behaves as `enable` and `false` behaves as `disable`.
</pre>

y-ken:
@hito4t 
Got it. After that, all of readme has these ssl option description.

```
$ grep ssl embulk-input-*/README.md
embulk-input-jdbc/README.md:- **ssl**: use SSL to connect to the database (string, default: `disable`. `enable` uses SSL without server-side validation and `verify` checks the certificate. For compatibility reasons, `true` behaves as `enable` and `false` behaves as `disable`.)
embulk-input-mysql/README.md:- **ssl**: use SSL to connect to the database (string, default: `disable`. `enable` uses SSL without server-side validation and `verify` checks the certificate. For compatibility reasons, `true` behaves as `enable` and `false` behaves as `disable`.
embulk-input-postgresql/README.md:- **ssl**: enables SSL. Data will be encrypted but CA or certification will not be verified (boolean, default: false)
embulk-input-redshift/README.md:- **ssl**: enables SSL. Data will be encrypted but CA or certification will not be verified (boolean, default: false)
```

y-ken:
@hito4t 
Got it. After that, all of readme has these ssl option description.

```
$ grep ssl embulk-input-*/README.md
embulk-input-jdbc/README.md:- **ssl**: use SSL to connect to the database (string, default: `disable`. `enable` uses SSL without server-side validation and `verify` checks the certificate. For compatibility reasons, `true` behaves as `enable` and `false` behaves as `disable`.)
embulk-input-mysql/README.md:- **ssl**: use SSL to connect to the database (string, default: `disable`. `enable` uses SSL without server-side validation and `verify` checks the certificate. For compatibility reasons, `true` behaves as `enable` and `false` behaves as `disable`.)
embulk-input-postgresql/README.md:- **ssl**: enables SSL. Data will be encrypted but CA or certification will not be verified (boolean, default: false)
embulk-input-redshift/README.md:- **ssl**: enables SSL. Data will be encrypted but CA or certification will not be verified (boolean, default: false)
```

hito4t:
@y-ken 
Thank you!

The type of the `ssl` option in embulk-input-postgresql and embulk-input-redshift is `boolean`, although the type of the `ssl` option in embulk-input-mysql is `enum Ssl`.
And because embulk-input-jdbc doesn't have `ssl` option, we should remove it from the readme.

y-ken:
oh, does it need to remove ssl option line from README for embulk-input-jdbc?

```
embulk-input-jdbc/README.md:- **ssl**: use SSL to connect to the database (string, default: `disable`. `enable` uses SSL without server-side validation and `verify` checks the certificate. For compatibility reasons, `true` behaves as `enable` and `false` behaves as `disable`.)
```

hito4t:
@y-ken 
Yes, neither `JdbcInputPlugin.GenericPluginTask` interface nor its super interface (`AbstractJdbcInputPlugin.InputPlugin`) have the `ssl` property.
Would you update README for embulk-input-jdbc?

y-ken:
@hito4t it seems some results, but it does not effect to use ssl. does it correct?
$ grep -r -i "ssl" ./embulk-input-jdbc/
./embulk-input-jdbc//src/main/java/org/embulk/input/jdbc/Ssl.java:public enum Ssl
./embulk-input-jdbc//src/main/java/org/embulk/input/jdbc/Ssl.java:    public static Ssl fromString(String value)
./embulk-input-jdbc//src/main/java/org/embulk/input/jdbc/Ssl.java:            throw new ConfigException(String.format("Unknown SSL value '%s'. Supported values are enable, true, disable, false or verify.", value));

hito4t:
@y-ken 
Thank you!

> it seems some results, but it does not effect to use ssl. does it correct?

`embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/Ssl.java` is a class shared among embulk-input-jdbc plugins.
But embulk-input-jdbc itself doesn't use `Ssl` class.

PR #174, opened 2020-01-16
Update ssl option doc README

y-ken:
@hito4t
follow-up for https://github.com/embulk/embulk-input-jdbc/pull/172

About the current readme of embulk-input-mysql.
In the case of `ssl: true` option. It results with this parameters.
```
jdbc:mysql://***:3306/*** options {useCompression=true, socketTimeout=1800000, useSSL=true, user=***, useLegacyDatetimeCode=false, requireSSL=true, tcpKeepAlive=true, verifyServerCertificate=false, useCursorFetch=true, connectTimeout=300000, password=***, zeroDateTimeBehavior=convertToNull}
```
From this part of readme, I recommend to use "nor" instead of "and". It can reduce miss-reading.
So, I''ll open this ticket.
```
**ssl**: use SSL to connect to the database (string, default: `disable`. `enable` uses SSL without server-side validation and `verify` checks the certificate.
```

current
image

proposal to change
image


hito4t:
@y-ken 
Thank you!

PR #178, opened 2020-03-11
Use the "org.embulk.embulk-plugins" Gradle plugin to build

dmikurube:
Along with that,

* Bump up the version to 0.11.0
* Upgrade the Gradle wrapper to use Gradle 6.2.2
* Upgrade depending embulk-core to 0.9.23
* Set JavaCompile option for deprecation
* Move default JDBC drivers to compileOnly
* Add default JDBC drivers in testCompile


dmikurube:
@hito4t Can you have a look at this and following PRs when you have time? Those are many changes, so we're not in a hurry.

Those are not to implement any new feature, but to catch up with the latest Embulk's plan with deprecations.

dmikurube:
Ugh, sorry, I found that generated gems with it are not working properly. I'm going to have a deeper look.

dmikurube:
@hito4t The gem problem is now fixed by `org.embulk.embulk-plugins:0.4.1`(https://github.com/embulk/gradle-embulk-plugins/pull/88). It's ready. Can you help taking a look when you have time?

dmikurube:
Rebased to follow #185.

dmikurube:
@hito4t Ah... I didn't expect building on Windows. Hmm, that `gemspec` would be uploaded to RubyGems.org, and that is used also from UNIX-like environments.

So, I don't think that escaped `\\` is the right way, instead we'll want to use `/` there even if building on Windows.

If you're okay, I'll try that fix in the Gradle plugin, and update this PR.

hito4t:
@dmikurube 

> So, I don't think that escaped \\ is the right way, instead we'll want to use / there even if building on Windows.

You’re right.
The current version uses `/` for file separator in `.gemspec` files even on Windows.

dmikurube:
@hito4t To address the Windows build issue, upgraded the Gradle plugin to v0.4.2 to introduce this fix https://github.com/embulk/gradle-embulk-plugins/pull/94 in 3a2d91086bed6a24c8db1ee76db414d313c3293e.

hito4t:
@dmikurube 
Thank you!

I was able to build gems on Windows.

The `test` task is executed when executing `gradlew gem` in the previous version, but not executed in the new version.
Is that intented change?

dmikurube:
@hito4t 

> The test task is executed when executing gradlew gem in the previous version, but not executed in the new version.
> Is that intented change?

Hmm, if `test` was executed for `./gradlew gem`, it was not intentional. The current behavior is correct.

If I understand the common Gradle manner correctly,
* `build` may run `test`.
* But `compileJava` nor `jar` does not run `test`.

dmikurube:
@hito4t Before this PR, `gem` was depending on `build`. I think that's why `gem` triggerred `test`.
https://github.com/embulk/embulk-input-jdbc/pull/178/files#diff-c197962302397baf3a4cc36463dce5eaL144

I can set `dependsOn` again if you want it to be configured like that although it's actually not a very common configuration in Gradle (ex. `jar` is not like that).

dmikurube:
Thanks! Let me merge the PRs #178, #179, #180, #181, #182, #183, and #184

PR #179, opened 2020-04-13
Replace Exec.getLogger with LoggerFactory.getLogger

dmikurube:
`Exec.getLogger` is now deprecated in Embulk. Plugins can call `LoggerFactory.getLogger` directly.

### DIFF
--- a/embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/PostgreSQLInputConnection.java
+++ b/embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/PostgreSQLInputConnection.java
@@ -6,7 +6,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import org.slf4j.Logger;
-import org.embulk.spi.Exec;
+import org.slf4j.LoggerFactory;
 import org.embulk.input.jdbc.JdbcInputConnection;
 import org.embulk.input.jdbc.JdbcLiteral;
 import org.embulk.input.jdbc.getter.ColumnGetter;
@@ -14,7 +14,7 @@ import org.embulk.input.jdbc.getter.ColumnGetter;
 public class PostgreSQLInputConnection
         extends JdbcInputConnection
 {
-    private final Logger logger = Exec.getLogger(PostgreSQLInputConnection.class);
+    private final Logger logger = LoggerFactory.getLogger(PostgreSQLInputConnection.class);
 
     public PostgreSQLInputConnection(Connection connection, String schemaName)
             throws SQLException
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -16,6 +16,7 @@ import java.sql.SQLException;
 import java.util.TreeMap;
 
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.base.Optional;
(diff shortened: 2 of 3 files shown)


embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/PostgreSQLInputConnection.java
```diff
@@ -6,15 +6,15 @@
...
+import org.slf4j.LoggerFactory;
 import org.embulk.input.jdbc.JdbcInputConnection;
 import org.embulk.input.jdbc.JdbcLiteral;
 import org.embulk.input.jdbc.getter.ColumnGetter;
 
 public class PostgreSQLInputConnection
         extends JdbcInputConnection
 {
-    private final Logger logger = Exec.getLogger(PostgreSQLInputConnection.class);
+    private final Logger logger = LoggerFactory.getLogger(PostgreSQLInputConnection.class);
```
hito4t, line 17: @dmikurube 
I think this line should be removed, because the `logger` instance is created in the superclass (`JdbcInputConnection`).

shroman: it looks valid to have two logger instances, doesn't it? ;) what's the concern here?

dmikurube: The `logger` instance in the superclass logs as `JdbcInputConnection`, and this `logger` instance in this class dumps logs as `PostgreSQLInputConnection`.

They have difference in dumped logs, and I believe the difference should be kept so that users/developers are aware which class dumps the log.

----

But at the same time, I also see that the `logger` instance in `JdbcInputConnection` is `protected` -- visible from `PostgreSQLInputConnection`. Their field names conflict. It may be okay because the `logger` instance of this class is `protected`, though... (I'm not very confident how Java language spec works for this case.)

Then, we may finally want to :
* Change the `logger` field of `JdbcInputConnection` to `private`.

What do you think? @hito4t

To keep this PR single-purposed, I'll do that in a different PR.

hito4t: @dmikurube 
> 
> The `logger` instance in the superclass dumps logs as `JdbcInputConnection`, and this `logger` instance in this class dumps logs as `PostgreSQLInputConnection`.
> 

The `logger` instance in the the `JdbcInputConnection` will dump log as the subclass, because  
`getClass()` returns a runtime class. 
(`JdbcInputConnection.logger = LoggerFactory.getLogger(getClass())`)

I verified that `JdbcInputConnection.logger` and `PostgreSQLInputConnection.logger` are same instance by Eclipse debugger.

If `JdbcInputConnection.logger` is `LoggerFactory.getLogger(JdbcInputConnection.class)`, it will dump log as the superclass(`JdbcInputConnection`).

dmikurube: @hito4t (Sorry to be much late.) Ah, got it. That's correct.

But honestly speaking, sharing a `protected` field variable (not a method) between a superclass and a subclass looks confusing to me, even if it's `final`. At the same time, it is common to maintain a `logger` instance as a class `static` field, not an instance field.

Then, what do you think about:
* Keep this in this PR. (I'm okay to keep it, but it will come again in the next step.)
* Then in another PR, make those `logger` instances `private static final` fields for each.

hito4t: @dmikurube 

>  Keep this in this PR. (I'm okay to keep it, but it will come again in the next step.)
>  Then in another PR, make those logger instances private static final fields for each.


Thank you, I agree with you.

dmikurube:
Thanks! I will file the other PR for the `logger` instance.

PR #181, opened 2020-04-13
Replace Joda-Time DateTimeZone to Java 8 java.time classes

dmikurube:
Along with making JRuby optional in Embulk, Joda-Time would be also deprecated. Java 8's `java.time` classes can replace them.

### DIFF
--- a/embulk-input-mysql/src/main/java/org/embulk/input/mysql/getter/AbstractMySQLTimestampIncrementalHandler.java
+++ b/embulk-input-mysql/src/main/java/org/embulk/input/mysql/getter/AbstractMySQLTimestampIncrementalHandler.java
@@ -9,22 +9,22 @@ import org.embulk.spi.Column;
 import org.embulk.spi.Exec;
 import org.embulk.spi.time.TimestampFormatter;
 import org.embulk.spi.time.TimestampParser;
-import org.joda.time.DateTimeZone;
 
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
+import java.time.ZoneId;
 import java.util.Optional;
 
 public abstract class AbstractMySQLTimestampIncrementalHandler
         extends AbstractIncrementalHandler
 {
-    protected final DateTimeZone sessionTimeZone;
+    protected final ZoneId sessionTimeZone;
     protected long epochSecond;
     protected int nano;
 
-    public AbstractMySQLTimestampIncrementalHandler(DateTimeZone sessionTimeZone, ColumnGetter next)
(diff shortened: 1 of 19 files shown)


embulk-input-mysql/src/main/java/org/embulk/input/mysql/getter/AbstractMySQLTimestampIncrementalHandler.java
```diff
@@ -9,22 +9,22 @@
...
 import java.sql.SQLException;
 import java.sql.Timestamp;
+import java.time.ZoneId;
 import java.util.Optional;
 
 public abstract class AbstractMySQLTimestampIncrementalHandler
         extends AbstractIncrementalHandler
 {
-    protected final DateTimeZone sessionTimeZone;
+    protected final ZoneId sessionTimeZone;
```
vietnguyen-td, line 23: I have a concern here, a `ZoneId` could have many `ZoneOffset`, should we use ZoneOffset in this?

hito4t:
> 
> 
> @hito4t Thanks for your comment. Good catch! It still fails, but during the run, not at the configuration step with `ConfigException`.
> 
> It should fail earlier at the configuration step. Changed to do so in: [b64e1e9](https://github.com/embulk/embulk-input-jdbc/commit/b64e1e9c536ea1ddf4d5bd4beb3e0f5579eb138e)

Time zone is not only for embulk-input-jdbc.
Many plugins will use time zone.
I think embulk-core instead of embulk-input-jdbc should check time zone in the config.

How about using `ZoneId` instead of `String` to represent time zone in the config, and creating serializer/deserializer for `ZoneId` in embulk-core?

dmikurube:
@hito4t That kind of special data conversion in `org.embulk.config.*` has made the config processing situation complicated in `embulk-core` (especially about dependency libraries and ClassLoaders).

So, the embulk-core team has decided not to support more conversions, and furthermore, deprecate some existing. Sorry for inconvenience.

hito4t:
@dmikurube 

I understand embulk-core can't check time zone.

I'm afraid other plugins would forget to check time zone, but I don't have a good idea.

By the way, could you add tests that have invalid time zone in the config?

dmikurube:
@hito4t (Sorry to be so late.) Thanks for your comment. Added a test in: 746761f0379c3c58413a23ba86e7a7dc3cf0846d

For the future, we're preparing a plugin-side library `embulk-util-config`, which is to process Embulk configs on plugin's side, replacing core-side `ConfigSource#loadConfig` and `TaskSource#loadTask`.
https://github.com/embulk/embulk-util-config

In this library, a plugin can configure its own Jackson SerDe and validation in it. Once the library is available widely, a similar validation would be available again running on plugin's side.
(But, we'll need to cleanup embulk-core's Jackson and around before that.)

dmikurube:
Thanks!

PR #182, opened 2020-04-13
Replace Embulk's Timestamp-related classes to embulk-util-timestamp and java.time

dmikurube:
Embulk's core has contained `TimestampFormatter` and `TimestampParser` in it, but they caused some compatibility problems in the past.

We are going to make them deprecated, and we're providing an independent Timestamp library `embulk-util-timestamp`. It does not depend on Embulk's core, nor JRuby.
https://github.com/embulk/embulk-util-timestamp

Once a plugin uses the Timestamp library, the plugin is not affected by any Timestamp changes in Embulk's core.

### DIFF
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/TimestampColumnGetter.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/TimestampColumnGetter.java
@@ -2,8 +2,8 @@ package org.embulk.input.jdbc.getter;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.time.Instant;
 import org.embulk.spi.PageBuilder;
-import org.embulk.spi.time.Timestamp;
 import org.embulk.spi.type.Type;
 import org.embulk.spi.type.Types;
 import org.embulk.util.timestamp.TimestampFormatter;
@@ -23,7 +23,7 @@ public class TimestampColumnGetter
     {
         java.sql.Timestamp timestamp = from.getTimestamp(fromIndex);
         if (timestamp != null) {
-            value = Timestamp.ofEpochSecond(timestamp.getTime() / 1000, timestamp.getNanos());
+            value = Instant.ofEpochSecond(timestamp.getTime() / 1000, timestamp.getNanos());
         }
     }
 
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/TimestampColumnGetter.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/TimestampColumnGetter.java
@@ -4,9 +4,9 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import org.embulk.spi.PageBuilder;
 import org.embulk.spi.time.Timestamp;
-import org.embulk.spi.time.TimestampFormatter;
 import org.embulk.spi.type.Type;
 import org.embulk.spi.type.Types;
+import org.embulk.util.timestamp.TimestampFormatter;
 
 public class TimestampColumnGetter
         extends AbstractTimestampColumnGetter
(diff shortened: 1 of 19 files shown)


embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/TimestampColumnGetter.java
```diff
@@ -23,7 +23,7 @@ protected void fetch(ResultSet from, int fromIndex) throws SQLException
     {
         java.sql.Timestamp timestamp = from.getTimestamp(fromIndex);
         if (timestamp != null) {
-            value = Timestamp.ofEpochSecond(timestamp.getTime() / 1000, timestamp.getNanos());
+            value = Instant.ofEpochSecond(timestamp.getTime() / 1000, timestamp.getNanos());
```
vietnguyen-td, line 26: you could use directly `value = timestamp.toInstant()`

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/TimestampWithTimeZoneIncrementalHandler.java
```diff
@@ -56,48 +55,20 @@ public JsonNode encodeToJson()
...
-        configSource.set("timezone", "UTC");
-        TimestampParser parser = new TimestampParser(
-            Exec.newConfigSource().loadConfig(ParserIntlTask.class),
-            configSource.loadConfig(ParserIntlColumnOption.class));
-
         try {
-            org.embulk.spi.time.Timestamp epoch = parser.parse(fromValue.asText());
-            Timestamp sqlTimestamp = new Timestamp(epoch.getEpochSecond() * 1000);
+            final Instant epoch = PARSER.parse(fromValue.asText());
+            Timestamp sqlTimestamp = new Timestamp(epoch.getEpochSecond() * 1000L);
```
vietnguyen-td, line 67: this could be `Timestamp.from(epoch)`

embulk-input-mysql/src/main/java/org/embulk/input/mysql/getter/MySQLDateTimeTimestampIncrementalHandler.java
```diff
@@ -38,10 +38,10 @@ public String getTimestampPattern()
     }
 
     @Override
-    public Timestamp utcTimestampToSessionTime(org.embulk.spi.time.Timestamp from)
+    public Timestamp utcTimestampToSessionTime(final Instant from)
     {
         // reconvert from UTC to session time_zone
-        final long reconverted = ZonedDateTime.ofInstant(Instant.ofEpochSecond(from.getEpochSecond()), ZoneOffset.UTC)
+        final long reconverted = ZonedDateTime.ofInstant(from, ZoneOffset.UTC)
```
vietnguyen-td, line 44: i think this code block qual to `return Timestamp.from(from.atZone(sessionTimeZone).toInstant());`

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/TimestampColumnGetter.java
dmikurube: Good catch!

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/TimestampWithTimeZoneIncrementalHandler.java
dmikurube: Good catch!

embulk-input-mysql/src/main/java/org/embulk/input/mysql/getter/MySQLDateTimeTimestampIncrementalHandler.java
dmikurube: Hmm, I don't think so. `from.atZone(sessionTimeZone).toInstant()` results in `Instant` same with `from`. The `withZoneSameLocal` method does the tricky computation.

PR #188, opened 2020-06-29
Declare a private static logger in each class

dmikurube:
As discussed in https://github.com/embulk/embulk-input-jdbc/pull/179#pullrequestreview-408056738, this PR is to have `private` `static` logger instances in each class.


dmikurube:
Thanks! Merging. Let's release v0.11.0, then.

PR #203, opened 2021-04-06
wrap UnknownHostException to ConfigException

vietnguyen-td:
We have a retry mechanism, when starting a job with an invalid configuration, it should not be retry.
This PR to wrap UnknownHostException to ConfigException when starting `transaction`

### DIFF
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -3,7 +3,6 @@ package org.embulk.input.jdbc;
 import java.io.File;
 import java.io.FileFilter;
 import java.net.MalformedURLException;
-import java.net.SocketTimeoutException;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.UnknownHostException;
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -3,8 +3,10 @@ package org.embulk.input.jdbc;
 import java.io.File;
 import java.io.FileFilter;
 import java.net.MalformedURLException;
+import java.net.SocketTimeoutException;
 import java.net.URISyntaxException;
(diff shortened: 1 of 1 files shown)


kietdo360 (approved):
LGTM

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -218,6 +220,10 @@ public ConfigDiff transaction(ConfigSource config,
             // TODO incremental_columns is not set => get primary key
             schema = setupTask(con, task);
         } catch (SQLException ex) {
+            final Throwable cause = ex.getCause();
+            if (cause instanceof SocketTimeoutException || cause instanceof UnknownHostException) {
```
dmikurube, line 224: It is understandable `UnknownHostException` to consider as `ConfigException`. But `SocketTimeoutException` does not sound like a `ConfigException` actually.

(Even when we consider a retry based on a type of Exception (sorry, it's our internal use-case, @hito4t), socket timeout is considered to recover by some retries?)

vietnguyen-td: When i am testing on local with invalid host, i also get both these exceptions with invalid `host`
- `UnknownHostException` when internet issue,  unable to resolve internet addresses
- `SocketTimeoutException` when server issue, fail to response

imo, those exceptions have chance to recover by some retries, we ignore only when start `transaction`

dmikurube: The problem here is NOT that you received those exceptions when testing on your local environments. The problem here is that real tentative network unstability is also (unintentionally or unexpectedly) affected by this.

vietnguyen-td: removed SocketTimeoutException

dmikurube (approved):
Okay, looks good to me!

@hito4t Unless you have any concern, I think we'll merge this early next week.

hito4t (approved):

ylzen (approved):
Reviewed for Data Exfiltrations.

dmikurube:
Resolved conflicts. Merging after confirming the CI is okay.

PR #205, opened 2021-04-10
Update GitHub Actions to v2 to fix build fail

vietnguyen-td:
The build starts failing when `set-env` is deprecated https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/. This PR to remove `set-env` and also update some config
- Change `env.workspace` to `github.workspace`
- Update GitHub action to v2
- Add test result in build log

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,6 +86,7 @@ jobs:
     - uses: actions/setup-java@v2
       with:
         java-version: 1.8
+        distribution: 'zulu'
     - name: Connect
       run: mysql -h 127.0.0.1 --port 3306 -uroot -proot -e "show databases;"
     - name: Create database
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -23,7 +23,11 @@ jobs:
           DBNAME: testdb
           SAMPLEDB: true
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
+    - uses: actions/setup-java@v2
+      with:
+        java-version: 8
+        distribution: 'zulu'
```
dmikurube, line 30: Does this mean Zulu build? Any reason to choose this? https://jp.azul.com/downloads/zulu-community/

dmikurube (approved):
Basically looks good (so approved), but left one question.

Also, added @hito4t in reviewers jfyi.

.github/workflows/build.yml
vietnguyen-td: Currently, `setup-java@v2` is just supporting two distributions `zulu & adopt` https://github.com/actions/setup-java

And `setup-java@v1` use `zulu` as default, so i keep it same as before

dmikurube: Ah, got it. `v1` was Zulu... Thanks!

PR #207, opened 2021-05-17
Catch up with Embulk API/SPI v0.10

dmikurube:
Getting these `embulk-input-jdbc` plugins caught up with Embulk v0.10+ API/SPI.

See also: https://dev.embulk.org/topics/get-ready-for-v0.11-and-v1.0.html

### DIFF
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/JdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/JdbcInputPlugin.java
@@ -7,11 +7,10 @@ import java.util.Properties;
 import java.sql.Connection;
 import java.sql.Driver;
 import java.sql.SQLException;
-import com.google.common.base.Throwables;
-import org.embulk.config.Config;
-import org.embulk.config.ConfigDefault;
 import org.embulk.input.jdbc.AbstractJdbcInputPlugin;
 import org.embulk.input.jdbc.JdbcInputConnection;
+import org.embulk.util.config.Config;
+import org.embulk.util.config.ConfigDefault;
 
 public class JdbcInputPlugin
         extends AbstractJdbcInputPlugin
@@ -81,8 +80,10 @@ public class JdbcInputPlugin
             // TODO check Class.forName(driverClass) is a Driver before newInstance
             //      for security
             driver = (Driver) Class.forName(t.getDriverClass()).newInstance();
-        } catch (Exception ex) {
-            throw Throwables.propagate(ex);
+        } catch (final RuntimeException | Error ex) {
+            throw ex;
+        } catch (final Exception ex) {
+            throw new RuntimeException(ex);
         }
 
         Connection con = driver.connect(t.getUrl(), props);
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/JdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/JdbcInputPlugin.java
@@ -80,7 +80,7 @@ public class JdbcInputPlugin
             // TODO check Class.forName(driverClass) is a Driver before newInstance
             //      for security
             driver = (Driver) Class.forName(t.getDriverClass()).newInstance();
-        } catch (final RuntimeException | Error ex) {
+        } catch (final RuntimeException ex) {
             throw ex;
         } catch (final Exception ex) {
             throw new RuntimeException(ex);
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -2,10 +2,15 @@ package org.embulk.input.jdbc;
 
 import java.io.File;
 import java.io.FileFilter;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.net.URLClassLoader;
 import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -21,18 +26,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.databind.JsonNode;
-import com.google.common.base.Throwables;
-import com.google.common.collect.ImmutableList;
 
-import org.embulk.config.Config;
 import org.embulk.config.ConfigException;
-import org.embulk.config.ConfigDefault;
 import org.embulk.config.ConfigDiff;
 import org.embulk.config.ConfigSource;
-import org.embulk.config.Task;
 import org.embulk.config.TaskReport;
 import org.embulk.config.TaskSource;
-import org.embulk.plugin.PluginClassLoader;
 import org.embulk.spi.BufferAllocator;
 import org.embulk.spi.Column;
 import org.embulk.spi.DataException;
@@ -45,6 +44,12 @@ import org.embulk.input.jdbc.getter.ColumnGetter;
 import org.embulk.input.jdbc.getter.ColumnGetterFactory;
 import org.embulk.input.jdbc.JdbcInputConnection.BatchSelect;
 import org.embulk.input.jdbc.JdbcInputConnection.PreparedQuery;
+import org.embulk.util.config.Config;
+import org.embulk.util.config.ConfigDefault;
+import org.embulk.util.config.ConfigMapper;
+import org.embulk.util.config.ConfigMapperFactory;
+import org.embulk.util.config.Task;
+import org.embulk.util.config.TaskMapper;
 
 import static java.util.Locale.ENGLISH;
 
@@ -52,6 +57,11 @@ public abstract class AbstractJdbcInputPlugin
         implements InputPlugin
 {
     private static final Logger logger = LoggerFactory.getLogger(AbstractJdbcInputPlugin.class);
+
+    protected static final ConfigMapperFactory CONFIG_MAPPER_FACTORY = ConfigMapperFactory.builder().addDefaultModules().build();
+
+    protected static final ConfigMapper CONFIG_MAPPER = CONFIG_MAPPER_FACTORY.createConfigMapper();
+    protected static final TaskMapper TASK_MAPPER = CONFIG_MAPPER_FACTORY.createTaskMapper();
 
     public interface PluginTask extends Task
     {
@@ -184,7 +194,7 @@ public abstract class AbstractJdbcInputPlugin
     public ConfigDiff transaction(ConfigSource config,
             InputPlugin.Control control)
     {
-        PluginTask task = config.loadConfig(getTaskClass());
+        final PluginTask task = CONFIG_MAPPER.map(config, this.getTaskClass());
 
         // Invalid timezones should fail immediately when configuring.
         throwAgainstInvalidTimeZone(task.getDefaultTimeZone());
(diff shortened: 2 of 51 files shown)


embulk-input-jdbc/src/main/java/org/embulk/input/JdbcInputPlugin.java
```diff
@@ -81,8 +80,10 @@ protected JdbcInputConnection newConnection(PluginTask task) throws SQLException
             // TODO check Class.forName(driverClass) is a Driver before newInstance
             //      for security
             driver = (Driver) Class.forName(t.getDriverClass()).newInstance();
-        } catch (Exception ex) {
-            throw Throwables.propagate(ex);
+        } catch (final RuntimeException | Error ex) {
```
vietnguyen-td, line 83: is `catch Error` needed because we just re-throw it again?

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -350,7 +360,7 @@ private String normalizeTableNameCase(JdbcInputConnection con, String tableName)
     private List<Integer> findIncrementalColumnIndexes(JdbcSchema schema, List<String> incrementalColumns)
         throws SQLException
     {
-        ImmutableList.Builder<Integer> builder = ImmutableList.builder();
+        final ArrayList<Integer> builder = new ArrayList<>();
```
vietnguyen-td, line 363: declare by interface is better? `final List<String> builder = new ArrayList<>()`

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -683,12 +693,37 @@ public boolean accept(File file) {
     protected void addDriverJarToClasspath(String glob)
     {
         // TODO match glob
-        PluginClassLoader loader = (PluginClassLoader) getClass().getClassLoader();
+        final ClassLoader loader = getClass().getClassLoader();
+        if (!(loader instanceof URLClassLoader)) {
+            throw new RuntimeException("Plugin is not loaded by URLClassLoader unexpectedly.");
+        }
+        if (loader.getClass().getCanonicalName().equals("org.embulk.plugin.PluginClassLoader")) {
```
vietnguyen-td, line 700: the old code expected `PluginClassLoader`, but the new one is not, is it intended?

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/ToStringMap.java
```diff
@@ -32,4 +21,17 @@ public Properties toProperties()
         props.putAll(this);
         return props;
     }
+
+    private static Map<String, String> mapToStringString(final Map<String, ToString> mapOfToString) {
+        final HashMap<String, String> result = new HashMap<>();
+        for (final Map.Entry<String, ToString> entry : mapOfToString.entrySet()) {
+            final ToString value = entry.getValue();
+            if (value == null) {
+                result.put(entry.getKey(), null);
```
vietnguyen-td, line 30: The old one is `null` string, the new is null, is it side effect somewhere?

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/FloatColumnGetter.java
```diff
@@ -43,7 +42,15 @@ public void longColumn(Column column)
         long l;
         try {
             // TODO configurable rounding mode
-            l = DoubleMath.roundToLong(value, RoundingMode.HALF_UP);
+            if (Math.getExponent(value) > Double.MAX_EXPONENT) {
```
vietnguyen-td, line 45: this look duplicated with https://github.com/embulk/embulk-input-jdbc/pull/207/files#diff-4dda1f846176bb53b2e92ac04a515c83f15193e310c295087a62fb7c515e5586R45, could we avoid it?

vietnguyen-td (commented):
general looks good, have few concerns

embulk-input-jdbc/src/main/java/org/embulk/input/JdbcInputPlugin.java
dmikurube: Ah, right. `Error` is not `Exception` while `RuntimeException` is `Exception`. Good catch. Will address this.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
dmikurube: I don't think so. It's just a local variable.

We may achieve some benefits from `List` if it's a signature of visible methods (argument or return type). But for just a local variable, there are no benefits.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
dmikurube: Ugh, it should be `if (!(loader.getClass()...))`. Thanks. Will fix it.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/ToStringMap.java
dmikurube: Ah, good catch! Will fix it.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/FloatColumnGetter.java
dmikurube: Hmm, let me consider having this as a method in `AbstractColumnGetter`.

dmikurube (commented):
Thanks for your comments! Will address them in a next commit.

embulk-input-jdbc/src/main/java/org/embulk/input/JdbcInputPlugin.java
dmikurube: Fixed at https://github.com/embulk/embulk-input-jdbc/pull/207/commits/bc8ebcb911f6dd3d817521b04142548f5fe5d228#diff-978621b46b8e1eb4b8eb4e8cea26c09e229e1bc405e1f8d998b1f69b481c413dR83

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
dmikurube: Fixed at https://github.com/embulk/embulk-input-jdbc/pull/207/commits/bc8ebcb911f6dd3d817521b04142548f5fe5d228#diff-052bbb667d3dda11eadb326e2a45faedac30c4ddf70e5492368d6205ad274a5cR700

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/ToStringMap.java
dmikurube: Fixed at https://github.com/embulk/embulk-input-jdbc/pull/207/commits/bc8ebcb911f6dd3d817521b04142548f5fe5d228#diff-4682fe5aa166d0303b1cde0cac83369807c974f61ff18f6981be2aa0c4bbc895R29

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/getter/FloatColumnGetter.java
dmikurube: Refactored out in https://github.com/embulk/embulk-input-jdbc/pull/207/commits/bc8ebcb911f6dd3d817521b04142548f5fe5d228#diff-a75524cd471aa6334d9b7362f0578902b884cc7ffd65bd9e7872e1b3812a8a9eR106-R130

But, they had a difference -- one receiving `float`, and the other receiving `double`. I thought about merging them to `double`, but kept as-is in `AbstractColumnGetter` fow now so that it does not introduce unnecessary impacts.

vietnguyen-td (approved):
LGTM

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -683,12 +693,37 @@ public boolean accept(File file) {
     protected void addDriverJarToClasspath(String glob)
     {
         // TODO match glob
-        PluginClassLoader loader = (PluginClassLoader) getClass().getClassLoader();
+        final ClassLoader loader = getClass().getClassLoader();
+        if (!(loader instanceof URLClassLoader)) {
+            throw new RuntimeException("Plugin is not loaded by URLClassLoader unexpectedly.");
+        }
+        if (!loader.getClass().getCanonicalName().equals("org.embulk.plugin.PluginClassLoader")) {
```
hito4t, line 700: `Class#getCanonicalName` may return `null`.
https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getCanonicalName--

Can we use `Class#getName` instead?

hito4t (commented):
@dmikurube 
I added some comments.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
hito4t: I think `builder` is not appropriate as `ArrayList` variable name (`indices` or `list`?).

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
dmikurube: Ah, thanks for your good catch. `null.equals(...)` would result in unexpected failure there. (It should fail anyway eventually, but it should fail in an expected way.)

I at first thought that a "canonical" name would sound better to confirm a kind of equality, but reading the Q&A below, I understood that `getName()` would be better.

> * the **name** is the name that you'd use to dynamically load the class with, for example, a call to `Class.forName` with the default `ClassLoader`. Within the scope of a certain `ClassLoader`, **all classes have unique names**.
> * the **canonical name** is the name that would be used in an import statement. It might be useful during `toString` or logging operations. ... (snip) ... thus **canonical names do not uniquely identify classes within a `ClassLoader`**.
>
> https://stackoverflow.com/questions/15202997/what-is-the-difference-between-canonical-name-simple-name-and-class-name-in-jav

Then in c56539e9f170b4a58ae3748b5e4cecaee2b763ff, I :
* changed it to `getName()`, and
* reversed the condition like below so that it works even if `getName()` returns an unexpected thing evem in the future.

```
       if (!"org.embulk.plugin.PluginClassLoader".equals(loader.getClass().getName())) {
```

What do you think?

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
dmikurube: Ah, in this PR, I focused to minimize the diff so that reviewers can focus on the essential difference. It has already been a big change, but it was impossible to split this PR into smaller portions.

(I also thought that it's still a kind of builder to build an "unmodifiable" list by `Collections.unmodifiableList` below, but I know it's a lame excuse...)

If you think it's not good, let's do a refactoring in another PR. What do you think?

hito4t: > let's do a refactoring in another PR. What do you think?

That's a good idea.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
hito4t: LGTM!

hito4t (approved):

dmikurube:
Thanks!

PR #213, opened 2021-05-28
Refactor variable names with "builder" for building unmodifiable collections

dmikurube:


hito4t (approved):

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -353,19 +353,19 @@ private String normalizeTableNameCase(JdbcInputConnection con, String tableName)
     private List<Integer> findIncrementalColumnIndexes(JdbcSchema schema, List<String> incrementalColumns)
         throws SQLException
     {
-        final ArrayList<Integer> builder = new ArrayList<>();
+        final ArrayList<Integer> indices = new ArrayList<>();
         for (String name : incrementalColumns) {
             Optional<Integer> index = schema.findColumn(name);
             if (index.isPresent()) {
```
vietnguyen-td, line 359: this could be shorten 
```
  indices.add(index.orElseThrow(() -> new ConfigException(String.format(ENGLISH,
          "Column name '%s' is in incremental_columns option does not exist",
          name))));
```

vietnguyen-td (approved):
minor comment about `Optional` class

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
dmikurube: Hmm, thought about that, but let's minimize the changes before releasing v0.12.0. Let's try a refactor after that.

dmikurube:
Thanks!

PR #220, opened 2021-08-12
Add statement timeout for PostgreSQL

hieudion:
`statement_timeout` is used quite useful in some cases. 
I would like to add this option to `embulk_input_postgresql`

### DIFF
--- a/embulk-input-postgresql/src/main/java/org/embulk/input/PostgreSQLInputPlugin.java
+++ b/embulk-input-postgresql/src/main/java/org/embulk/input/PostgreSQLInputPlugin.java
@@ -54,9 +54,9 @@ public class PostgreSQLInputPlugin
         @ConfigDefault("\"embulk-input-postgresql\"")
         public String getApplicationName();
 
-        @Config("statement_timeout")
-        @ConfigDefault("-1")
-        public int getStatementTimeout();
+        @Config("statement_timeout_millis")
+        @ConfigDefault("null")
+        public Optional<Integer> getStatementTimeoutMillis();
     }
 
     @Override
@@ -100,7 +100,7 @@ public class PostgreSQLInputPlugin
 
         Connection con = DriverManager.getConnection(url, props);
         try {
-            PostgreSQLInputConnection c = new PostgreSQLInputConnection(con, t.getSchema(), t.getStatementTimeout());
+            PostgreSQLInputConnection c = new PostgreSQLInputConnection(con, t.getSchema(), t.getStatementTimeoutMillis().orElse(-1));
             con = null;
             return c;
         } finally {
--- a/embulk-input-postgresql/src/main/java/org/embulk/input/PostgreSQLInputPlugin.java
+++ b/embulk-input-postgresql/src/main/java/org/embulk/input/PostgreSQLInputPlugin.java
@@ -100,7 +100,7 @@ public class PostgreSQLInputPlugin
 
         Connection con = DriverManager.getConnection(url, props);
         try {
-            PostgreSQLInputConnection c = new PostgreSQLInputConnection(con, t.getSchema(), t.getStatementTimeoutMillis().orElse(-1));
+            PostgreSQLInputConnection c = new PostgreSQLInputConnection(con, t.getSchema(), t.getStatementTimeoutMillis());
             con = null;
             return c;
         } finally {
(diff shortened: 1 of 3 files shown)


embulk-input-postgresql/src/main/java/org/embulk/input/PostgreSQLInputPlugin.java
```diff
@@ -53,6 +53,10 @@
         @Config("application_name")
         @ConfigDefault("\"embulk-input-postgresql\"")
         public String getApplicationName();
+
+        @Config("statement_timeout")
+        @ConfigDefault("-1")
+        public int getStatementTimeout();
```
dmikurube, line 59: `-1` to disable (by default) seems like a traditional manner. How about:

```
@Config("statement_timeout")
@ConfigDefault("null")
public Optional<Integer> getStatementTimeout();
```

Even if we use a number to disable, `0` won't work?

----

At the same time, I was at first concerned about its unit (seconds or milliseconds), and thought to name it `statement_timeout_millis` or like that.

But, for the consistency with the existing `connect_timeout` and `socket_timeout` (they look to be milliseconds), `statement_timeout` would be good.

embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/PostgreSQLInputConnection.java
```diff
@@ -76,4 +78,19 @@ public void close() throws SQLException
             // TODO close?
         }
     }
+
+    private void setStatementTimeout(int statementTimeout)
```
dmikurube, line 82: It does not always set the statement timeout. How about to name it :

```
private void setStatementTimeoutIfSpecified(Optional<Integer> statementTimeout)
```

embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/PostgreSQLInputConnection.java
```diff
@@ -76,4 +78,19 @@ public void close() throws SQLException
...
+        throws SQLException
+    {
+        if (statementTimeout > -1) {
+            Statement stmt = connection.createStatement();
+            try {
+                String sql = "SET statement_timeout TO " + quoteIdentifierString(String.valueOf(statementTimeout));
```
dmikurube, line 88: Looking at other SQL methods (e.g. `newBatchSelect` above), it could be nice to have a log like:

```
logger.info("SQL: " + sql);
```

embulk-input-redshift/src/main/java/org/embulk/input/RedshiftInputPlugin.java
```diff
@@ -46,6 +46,10 @@
         @Config("ssl")
         @ConfigDefault("false")
         public boolean getSsl();
+
+        @Config("statement_timeout")
+        @ConfigDefault("-1")
+        public int getStatementTimeout();
```
dmikurube, line 52: Same with PostgreSQL.

embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/PostgreSQLInputConnection.java
```diff
@@ -76,4 +78,19 @@ public void close() throws SQLException
             // TODO close?
         }
     }
+
+    private void setStatementTimeout(int statementTimeout)
+        throws SQLException
+    {
+        if (statementTimeout > -1) {
```
dmikurube, line 85: Based on my suggestions above, how about:
```
if (statementTimeout.isPresent() && statementTimeout.get() > 0) {
```

embulk-input-postgresql/src/main/java/org/embulk/input/PostgreSQLInputPlugin.java
hieudion: if we set it to 0 it means no timeout.

Thanks for your suggestion.

dmikurube (approved):
Looks good!

@hito4t Can we have your look?

dmikurube:
JFYI, `embulk-input-sqlserver`'s checks are fixed separete in #221.

hito4t (approved):
LGTM!

dmikurube:
Thanks!

yasutaka-emoto:
Hi, Could you tell me what parameter name we should set in yaml file to change statement timeout of postgres ? Is is correct "setStatementTimeout" and we have to set value as millisecond ?

hiroyuki-sato:
Hello, @yasutaka-emoto 

Could you try the following?

* embulk-input-postgresql: 0.12.3
* `statement_timeout_millis`

https://github.com/embulk/embulk-input-jdbc/pull/220/files#diff-6058ddaf7edb070082c1f68aa2f481a21a7cbfb072dbfb3eae2bf28dbace418eR57

yasutaka-emoto:
@hiroyuki-sato Hi, Thank you for your quick replying ! I will try it to check whether it works.

PR #224, opened 2021-08-24
add query timeout options

case-k-git:
add query timeout options 

```
queryTimeoutint-1 | The number of seconds to wait before a timeout has occurred on a query. The default value is -1, which means infinite timeout. Setting this value to 0 also implies to wait indefinitely.
-- | --
```

https://docs.microsoft.com/en-us/sql/connect/jdbc/setting-the-connection-properties?redirectedfrom=MSDN&view=sql-server-ver15

### DIFF
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -135,6 +135,9 @@ public abstract class AbstractJdbcInputPlugin
         //@Config("limit_rows")
         //@ConfigDefault("null")
         //public Optional<Integer> getLimitRows();
+        @Config("query_timeout")
+        @ConfigDefault("-1")
+        public int getQueryTimeout();
 
         @Config("connect_timeout")
         @ConfigDefault("300")
--- a/embulk-input-mysql/README.md
+++ b/embulk-input-mysql/README.md
@@ -37,6 +37,7 @@ MySQL input plugin for Embulk loads records from MySQL.
   - If this value is set to -1:
     - It uses a client-side built statement and fetches all rows at once. This may cause OutOfMemoryError.
     - Internally, `useCursorFetch=false` is used and `java.sql.Statement.setFetchSize` is not set.
+- **query_timeout**: timeout for query connect. -1 means no timeout. (integer (seconds), default: -1)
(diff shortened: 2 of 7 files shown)


hiroyuki-sato:
Hello, @case-k-git 

I'm not an SQLServer user.
Have you ever tried the following configuration?
It seems that this configuration works the same.
Could you tell me the detail of why do you need to add this option?

```yaml
in:
  type: sqlserver
  options:
    queryTimeout: 5
    # ... 
```

case-k-git:
Hello @hiroyuki-sato

thank you for your check.

>Have you ever tried the following configuration?
 I see ... I did not notice that options . I gonna use it . but could you please wait until I check?
 I have tried to use queryTimeout options. but did not timeout as I expected both options one and my fixed one.

>Could you tell me the detail of why do you need to add this option?

I would like to avoid keep executing unintentional query not to press database disk.
(please check the detail under the tech blog about snapshot separation)
https://techblog.zozo.com/entry/sqlserver-transaction-isolation-level-snapshot
in our company we are using  snapshot separation to avoid with(nolock) data lost and data duplicate issue.

by using snapshot the query gonna be write into the tempdb and if SELECT Query is keep executed tempdb disk gonna press and cause disk issue. 

to avoid the risk of this we wanna set the query timeout  option not to execute unintentional query

thank you

hiroyuki-sato:
Hello, @case-k-git 

This is just in my(as a this plugin user) opinion,

For example, I used some MySQL-specific JDBC options in the embulk-input-mysql plugin.
I think you can use the same option in the embulk-input-sqlserver plugin. So, please try it.
If the option doesn't work correctly, It needs to fix the problem at first.

```yaml
in:
  type: mysql
  # mysql
  options: { useLegacyDatetimeCode: false, serverTimezone: "Pacific/Midway", noTimezoneConversionForDateType: false }
```

You can check the JDBC option in the logs like the below.

```yaml
in:
  type: mysql
  options:
    hoge: HOGE
    fuga: FUGA
```

```
2021-08-25 17:41:15.154 +0900 [INFO] (0001:transaction): Connecting to jdbc:mysql://localhost:3306/embulk_test options {useCompression=true, socketTimeout=1800000, useSSL=false, user=root, useLegacyDatetimeCode=false, tcpKeepAlive=true, hoge=HOGE, fuga=FUGA, useCursorFetch=true, connectTimeout=300000, password=***, zeroDateTimeBehavior=convertToNull}
```

And If this plugin supports the new `query_timeout` option, I hope that embulk-output-sqlsever also supports the same option.(If this option is useful for an output plugin.)

PR #230, opened 2022-05-31
Add adal4j to support active directory authentication

legiangthanh:
mssql-jdbc will dynamically load the ada4j library when authentication is `ActiveDirectoryPassword`. But currently, the gem's classpath doesn't contain the ada4j library, mssql-jdbc will throw an exception.

```
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: Failed to load ADAL4J Java library for performing ActiveDirectoryPassword authentication.
        at com.microsoft.sqlserver.jdbc.SQLServerConnection.validateAdalLibrary(SQLServerConnection.java:4233)
```

This PR will add the add4j library to the dependency of sqlserver to include this library in the gem's classpath.


dmikurube (commented):
It basically looks good, but can you leave an explicit technical background description in the PR description how "Just adding the `adal4j` library makes it support the ACtive Directory-based authentication"?

dmikurube (approved):
Thanks. Okay, the description looks reasonable.

PR #233, opened 2022-06-30
Use the MySQL/PostgreSQL JDBC driver found in the class loader in the highest priority

dmikurube:
The JDBC plugins will be available as Maven-based plugins. To make it realized, their `pom.xml` should include MySQL/PostgreSQL JDBC drivers so that Embulk's `PluginClassLoader` can find the JDBC drivers from `pom.xml`. Then, they are loaded straightforward from the class loader.

But currently, their driver finding mechanism does not prioritize classes found straightforward from the class loader.

This pull request changes the priority, including some refactoring with exception/log message changes.


vietnguyen-td (approved):
LGTM

dmikurube:
You mean some failures in `checkstyle`? If so, yeah, I also noticed, but they're configured non-critical. We may want to address them later...

dmikurube:
Thanks!

PR #234, opened 2022-07-05
Use the Gradle plugin v0.5.2 with including default JDBC drivers as "optional"

dmikurube:
It enables MySQL and PostgreSQL plugins to be loaded in the Maven form, not the RubyGem form.


vietnguyen-td (approved):
LGTM

dmikurube:
Thanks!

PR #235, opened 2022-07-05
Move packages of plugin classes from org.embulk.input

dmikurube:
Java 11+ "modules" do not allow that multiple JAR files provide the same Java "package".

Historically, the plugin classes of all JDBC Input plugins have been in `org.embulk.input`, not `org.embulk.input.jdbc`, `org.embulk.input.mysql`, ... `JdbcInputPlugin` has been in `org.embulk.input`, `MySQLInputPlugin` has been also in `org.embulk.input`, `PostgreSQLInputPlugin` has been also in `org.embulk.input`, ...

Therefore, the JDBC plugins can have troubles to work with Java 11+.

This pull request is to get it fixed although it can be incompatible with other plugins that depend on JDBC plugins.

### DIFF
--- a/embulk-input-jdbc/build.gradle
+++ b/embulk-input-jdbc/build.gradle
@@ -2,7 +2,7 @@ dependencies {
 }
 
 embulkPlugin {
-    mainClass = "org.embulk.input.JdbcInputPlugin"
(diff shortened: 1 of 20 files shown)


vietnguyen-td (approved):
LGTM

dmikurube:
Thanks!

PR #246, opened 2022-11-21
Get last_record from TaskReport as List

dmikurube:
We observed a corner-case problem with Embulk v0.10.37 (maybe v0.10.33+?) that it cannot get `last_record` from `TaskReport` correctly.

```
  Caused by: org.embulk.config.ConfigException: com.fasterxml.jackson.databind.JsonMappingException: Can not construct instance of com.fasterxml.jackson.databind.JsonNode, problem: abstract types either need to be mapped to concrete types, have custom deserializer, or be instantiated with additional type information
   at [Source: N/A; line: -1, column: -1]
  	at org.embulk.util.config.DataSourceImpl.get(DataSourceImpl.java:75) ~[na:na]
  	at org.embulk.input.jdbc.AbstractJdbcInputPlugin.buildNextConfigDiff(AbstractJdbcInputPlugin.java:428) ~[na:na]
  	at org.embulk.input.jdbc.AbstractJdbcInputPlugin.transaction(AbstractJdbcInputPlugin.java:230) ~[na:na]
```

It looks coming from Embulk's change that the core has its own Jackson, and plugins have their own Jackson.

This pull-request may or may not resolve the problem. It's fine if it resolves the problem. It logs some details even if it does not resolve the problem.

### DIFF
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -30,6 +30,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import org.embulk.config.ConfigException;
 import org.embulk.config.ConfigDiff;
 import org.embulk.config.ConfigSource;
+import org.embulk.config.DataSource;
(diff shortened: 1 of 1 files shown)


vietnguyen-td (approved):
LGTM

dmikurube:
Thanks!

PR #251, opened 2023-06-20
support preview data

joe-td:
Currently, the embulk input jdbc connector fetch all records for preview. This PR will limit to 50 records for generating preview.

### DIFF
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
@@ -17,7 +17,6 @@ import java.util.Properties;
 import java.nio.file.Paths;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.time.DateTimeException;
 import java.time.ZoneId;
 import java.util.Optional;
 import java.util.TreeMap;
@@ -501,7 +500,8 @@ public abstract class AbstractJdbcInputPlugin
             }
 
             List<ColumnGetter> getters = newColumnGetters(con, task, querySchema, pageBuilder);
-            try (BatchSelect cursor = con.newSelectCursor(builtQuery, getters, task.getFetchRows(), task.getSocketTimeout())) {
+            try (BatchSelect cursor = con.newSelectCursor(builtQuery, getters, task.getFetchRows(),
+                task.getSocketTimeout(), Exec.isPreview())) {
                 while (true) {
                     long rows = fetch(cursor, getters, pageBuilder);
                     if (rows <= 0L) {
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/PreviewQueryBuilder.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/PreviewQueryBuilder.java
@@ -1,0 +1,41 @@
+package org.embulk.input.jdbc;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.regex.Pattern;
+
+public class PreviewQueryBuilder
+{
+    private static final String MSSQL_PRODUCT_NAME = "Microsoft SQL Server";
+    private static final String LIMIT_STATEMENT_TEMPLATE = "%s LIMIT 50";
+    private static final String TOP_STATEMENT_TEMPLATE = "%s TOP 50 %s";
+    private static final Pattern SELECT_PATTERN = Pattern.compile("select", Pattern.CASE_INSENSITIVE);
+
+    private String query;
+    private Connection connection;
+
+    public PreviewQueryBuilder(final String query, final Connection connection)
+    {
+        this.query = query;
+        this.connection = connection;
+    }
+
+    public String build() throws SQLException
+    {
+        return isMSSQL()
+            ? buildPreviewSQLForMSSQL()
+            : String.format(LIMIT_STATEMENT_TEMPLATE, this.query);
+    }
+
+    private String buildPreviewSQLForMSSQL()
+    {
+        final String[] statements = SELECT_PATTERN.split(query);
+        return String.format(TOP_STATEMENT_TEMPLATE, "SELECT", statements[1]);
+    }
+
+    protected boolean isMSSQL() throws SQLException
+    {
+        final String productName = connection.getMetaData().getDatabaseProductName();
+        return MSSQL_PRODUCT_NAME.equals(productName);
+    }
+}
--- a/embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/PostgreSQLInputConnection.java
+++ b/embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/PostgreSQLInputConnection.java
@@ -8,6 +8,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.Optional;
 
+import org.embulk.input.jdbc.PreviewQueryBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.embulk.input.jdbc.JdbcInputConnection;
@@ -27,11 +28,15 @@ public class PostgreSQLInputConnection
     }
 
     @Override
-    protected BatchSelect newBatchSelect(PreparedQuery preparedQuery,
-            List<ColumnGetter> getters,
-            int fetchRows, int queryTimeout) throws SQLException
+    protected BatchSelect newBatchSelect(PreparedQuery preparedQuery, List<ColumnGetter> getters, int fetchRows,
+                                         int queryTimeout, boolean isPreview) throws SQLException
     {
-        String query = "DECLARE cur NO SCROLL CURSOR FOR " + preparedQuery.getQuery();
+        String query = isPreview
+            ? new PreviewQueryBuilder(preparedQuery.getQuery(), connection).build()
+            : preparedQuery.getQuery();
+
+        query = "DECLARE cur NO SCROLL CURSOR FOR " + query;
+
         List<JdbcLiteral> params = preparedQuery.getParameters();
 
         logger.info("SQL: " + query);
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
@@ -123,18 +123,19 @@ public class JdbcInputConnection
         }
     }
 
-    public BatchSelect newSelectCursor(PreparedQuery preparedQuery,
-            List<ColumnGetter> getters,
-            int fetchRows, int queryTimeout) throws SQLException
+    public BatchSelect newSelectCursor(PreparedQuery preparedQuery, List<ColumnGetter> getters, int fetchRows,
+                                       int queryTimeout, boolean isPreview) throws SQLException
     {
-        return newBatchSelect(preparedQuery, getters, fetchRows, queryTimeout);
+        return newBatchSelect(preparedQuery, getters, fetchRows, queryTimeout, isPreview);
     }
 
-    protected BatchSelect newBatchSelect(PreparedQuery preparedQuery,
-            List<ColumnGetter> getters,
-            int fetchRows, int queryTimeout) throws SQLException
+    protected BatchSelect newBatchSelect(PreparedQuery preparedQuery, List<ColumnGetter> getters, int fetchRows,
+                                         int queryTimeout, boolean isPreview) throws SQLException
     {
-        String query = preparedQuery.getQuery();
+        String query = isPreview
+            ? new PreviewQueryBuilder(preparedQuery.getQuery(), connection).build()
+            : preparedQuery.getQuery();
+
         List<JdbcLiteral> params = preparedQuery.getParameters();
 
         PreparedStatement stmt = connection.prepareStatement(query);
--- a/embulk-input-mysql/src/main/java/org/embulk/input/mysql/MySQLInputConnection.java
+++ b/embulk-input-mysql/src/main/java/org/embulk/input/mysql/MySQLInputConnection.java
@@ -10,6 +10,7 @@ import java.util.TimeZone;
 
 import org.embulk.input.jdbc.JdbcInputConnection;
 import org.embulk.input.jdbc.JdbcLiteral;
+import org.embulk.input.jdbc.PreviewQueryBuilder;
 import org.embulk.input.jdbc.getter.ColumnGetter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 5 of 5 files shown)


embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/PreviewQueryBuilder.java
```diff
@@ -0,0 +1,41 @@
+package org.embulk.input.jdbc;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.regex.Pattern;
+
+public class PreviewQueryBuilder
+{
+    private static final String MSSQL_PRODUCT_NAME = "Microsoft SQL Server";
```
dmikurube, line 9: It does not seem very reasonable to have MSSQL-specific logic in this generic `PreviewQueryBuilder`. You may want to have a MSSQL-specific query builder in `embulk-input-sqlserver`, and use it there?  Then it does not require `connection` as well.

Or, it may apply also for others, MySQL and PostgreSQL. Even if their implementation would be the same, you may want to have a query builder for MySQL in `embulk-input-mysql`, and another query builder for PostgrSQL in `embulk-input-postgresql`. The single responsibility principle is often more prioritized than DRY.

dmikurube (commented):
Some comments.

1. Why 50?  Do you have any specific reason why 50 is good there?  Or, shouldn't it be user-configurable?
2. The PR subject "support preview data" does not seem descriptive enough. They already "support" previews. Give a better descriptive subject.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/PreviewQueryBuilder.java
joe-td: > Why 50? Do you have any specific reason why 50 is good there? Or, shouldn't it be user-configurable?
Actually, I don't know which number is best, so just pick random 50 records for preview is enough. It's better to move it to configurable, but I don't think we need it because it's just for display some sample record in our console UI.

joe-td:
> 1. Why 50?  Do you have any specific reason why 50 is good there?  Or, shouldn't it be user-configurable?

Actually, I don't know which number is best, so just pick random 50 records for preview is enough. It's better to move it to configurable, but I don't think we need it because it's just for display some sample record in our console UI.

dmikurube:
> I don't think we need it because it's just for display some sample record in our console UI.

Who are "our" ?  Please be aware that this `embulk-input-jdbc` is a public open source project. It's not "owned" by you, nor by your company. It is owned by the Embulk project community.

The Embulk team is unaware of your context. Please make a commit that is reasonable for all users of `embulk-input-jdbc` in the world, not only for you nor for your company. The same for the pull request subject and description.

joe-td:
> > I don't think we need it because it's just for display some sample record in our console UI.
> 
> Who are "our" ? Please be aware that this `embulk-input-jdbc` is a public open source project. It's not "owned" by you, nor by your company. It is owned by the Embulk project community.
> 
> The Embulk team is unaware of your context. Please make a commit that is reasonable for all users of `embulk-input-jdbc` in the world, not only for you nor for your company. The same for the pull request subject and description.

Got your point and noted. Let discuss about the number of limit record. In my opinion hard code number is enough. If the connector needs to get it from config, we must pass config to class `XXXInputConnection`.

hiroyuki-sato:
Hello, @joe-td 

I have one concern. Does this extension support incremental loading and raw query?
If not, does this PR skip limitation in those cases?
https://github.com/embulk/embulk-input-jdbc/tree/master/embulk-input-jdbc#use-incremental-loading-with-raw-query

joe-td:
Hi @hiroyuki-sato 
This patch is only impact on `embulk preview ...` command, so there is no issue with incremental loading. About raw query, I think it can work. It's just rebuild the query by append limit/top to original query.

hiroyuki-sato:
Hello, @joe-td 

A user can write any query like this. Does this PR append `limit/top` statement if a user use `query` attribute?
Or does this PR append `limit/top` statement if user uses `table` attribute only?

```sql
in:
  type: postgresql
  user: user
  host: localhost
  database: embulk_test
#  table: sample
  query: select * from sample limit 1
```

If a user uses `limit` statement in the `query` attribute, it may cause like a following error. 

```
select * from sample limit 1 limit 1;
ERROR:  syntax error at or near "limit"
LINE 1: select * from sample limit 1 limit 1;
```

joe-td:
Hi @hiroyuki-sato 
Yes, you're right. If user input raw query and the query contain top/limit, the current logic still append it and cause a problem.

hiroyuki-sato:
Hello, @joe-td 

I think `preview_query` is a little complicated if the user use `table` attribute only.

Just for one idea.
* `preview_limit`: If a configuration contains a `table` attribute, it appends `limit <preview_limit>` at the end of the query in the preview mode.
* `preview_query`: A configuration containing a `query` attribute uses `preview_query` in the preview mode. If the configuration doesn't have `preview_query`, It uses the `query` attribute as is.

Plan B. Introduce limit place mark ie.`<limit_stmt>` .

```yaml
in:
  type: postgresql
  user: user
  host: localhost
  database: embulk_test
#  table: sample
  query: select * from sample <limit_stmt>
```

joe-td:
Thanks for your suggestion. Actually I don't like to add more configuration for user, it can put user to spend more time on correct the config and easier to make error. One more thing about the raw query is about we don't know the best place to put it into the original query. I checked the JDBC driver and found out that they provide the method `setMaxRows()` https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#setMaxRows-int-, I think we can use this without introducing configuration or manipulate the query.

hiroyuki-sato:
Hello, @joe-td.

If `setMaxRows()` works as you expected, the implementation will be more straightforward. 👍

hiroyuki-sato:
Hello, @joe-td. Thank you for improving the PR.

The Embulk project is carefully keeping the current behavior as far as possible.
So, I'm worrying about changing the default preview limit suddenly.

Keeping current behavior and changing behavior using configuration may be helpful.
Fortunately, Embulk 0.11.0 introduce [EEP-4 System-wide Configuration by Embulk System Properties and Embulk Home](https://github.com/embulk/embulk/blob/master/docs/eeps/eep-0004.md). If Saas provider sets preview value globally using this setting, a user doesn't worry about preview size.

This is just my opinion.  I'll support @dmikurbe and @hito4t opinions.

dmikurube:
The configuration would be similar to `preview_sample_buffer_bytes` for File Input Plugins, and I found `preview_sample_rows` in the core.

https://github.com/embulk/embulk/blob/v0.11.0/embulk-core/src/main/java/org/embulk/exec/PreviewExecutor.java#L32-L58

I do not know actual example cases that Input Plugins use this configuration explicitly, but it might help somehow, or it would provide some hints.

joe-td:
I think the constant `MAX_PREVIEW_RECORDS` is just limit number of record for `embulk preview ...` and it avoid to scan all records from table to make query a bit fast for preview and it does not impact to `embulk run ...`. After the records is fetch from database, it will use the preview processor to display preview the data. In my opinion the purpose of `embulk preview` is display some sample data, so I don't see any risk at the moment.

dmikurube:
I understand your point, but for example, if there is a user who has used `embulk preview` to check 100+ records, the user will not have an option to keep usint it. Compatibility is not always mandatory, but we need an appropriate communication when breaking the compatibility, and we should avoid breaking the compatibility unless it's **very** difficult. In addition, even if we change the default behavior, we should leave an option to achieve compatible behavior if possible. In this case, it's limited to `100` and there is no way to avoid. It's not good.

Unfortunately, there seem no ways to pass the core's `preview_sample_rows` to an Input Plugin. (It could be a good point to improve the core side, but it's an independent topic.)

My suggestion is:
* Add a new configuration `preview_sample_rows` for these plugins. There would be no limit if it's `null`. The default would be `null`, but nice to print a warning log message that the preview query has no limit.
* You could have `OptionalInt limit`, instead of `boolean isPreview`, as the new parameter of `newSelectCursor`. The code changes could be easier, and more visible.

joe-td:
Thanks for your feedback and suggestion. Totally agree with you about the make it configurable. But currently behavior is the plugin print only 127 records for the samples even if we does not limit the records fetch from DB. There are 2 thing that we need to make it clear:
- Limit number of row fetch from DB for preview flow (currently there is no limit, it will scan all rows but fetch only 10000 records from cursor via `fetch_rows` config attribute, but I don't know it prints only 127 lines)
- Limit number of line print for preview ( current it config to 15 lines)

dmikurube:
Okay, then can you investigate why it prints only 127 lines?  Depending on the investigation, we may agree that limiting by default in preview would be reasonable.

dmikurube:
Jfyi, filed this for the core. https://github.com/embulk/embulk/issues/1623

joe-td:
I think the embulk internally handle page by page with limit page size is 32768 bytes. So, the number of preview lines will be very and depend on size of record.

- https://github.com/embulk/embulk/blob/v0.11.0/embulk-core/src/main/java/org/embulk/exec/PooledBufferAllocator.java#L46
- https://github.com/embulk/embulk-input-jdbc/blob/master/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java#L491
- https://github.com/embulk/embulk/blob/v0.11.0/embulk-core/src/main/java/org/embulk/exec/PreviewExecutor.java#L175

dmikurube:
Yeah, the internal `Page`/`Buffer` is limited to 32768, but it's the same in `run`, not just `preview`. Can you elaborate why it's consequently limited only in preview?

joe-td:
In the preview it just handle first page only: https://github.com/embulk/embulk/blob/v0.11.0/embulk-core/src/main/java/org/embulk/exec/PreviewExecutor.java#L175 and skip the remaining page.

dmikurube:
That is just checking the number of records (rows), not pages?

joe-td:
As my understand, it consumed first page (32678 bytes) and check number of rows from first page with default sample row config ?

dmikurube:
`PageOutput` is used repeatedly for multiple `Page`s. `recordCount` can be accumulated among multiple `Page`s. It look to me that `SamplingPageOutput` would accept the second `Page` if the first `Page`'s record count (by `PageReader#getRecordCount(Page)`) is less than `sampleRows`.

If you're showing it's limited to a single `Page`, you'll have to confirm `SamplingPageOutput#add(Page)` is called only once.

joe-td:
Yeah! the `SamplingPageOutput` can accept more than one page and after processing the page, the `recordCount` will accumulate and check with sample row config (15 lines). Back to my example, the first page has more than 15 records already, so it skip to subsequence page.

dmikurube:
So, the limit comes from `sampleRow`, which comes from the global `preview_sample_rows`.

Then let's get back. If there is a user who has used embulk preview to check 100+ records, the user would set the global `preview_sample_rows` to 100+.

But, your change will not work for this case. Even if the user sets 100+ for the global `preview_sample_rows`, the preview result would always be limited to 100. The user would have no any option to avoid it.

joe-td:
Thanks @dmikurube. It makes sense now. Let me add `preview_sample_rows` config attribute for this case. But I think it's better to transfer `preview_sample_rows` from core to plugin.

dmikurube:
> But I think it's better to transfer `preview_sample_rows` from core to plugin.

Agreed, but it'd be a long journey as it needs to change the contract between the core and plugins. If you need to limit this sooner, this is the only way.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -501,7 +505,8 @@ public TaskReport run(TaskSource taskSource,
             }
 
             List<ColumnGetter> getters = newColumnGetters(con, task, querySchema, pageBuilder);
-            try (BatchSelect cursor = con.newSelectCursor(builtQuery, getters, task.getFetchRows(), task.getSocketTimeout())) {
+            try (BatchSelect cursor = con.newSelectCursor(builtQuery, getters, task.getFetchRows(),
+                task.getSocketTimeout(), task.getPreviewSampleRows())) {
```
dmikurube, line 509: By this, it seems like `preview_sample_rows` always affects, even if it's not in the preview mode. Checking `Exec.isPreview()` in each `*InputConnection` does not look reasonable to me.

I meant that : 
* `newSelectCursor` would receive `OptionalInt limit` in general for any case, either in the preview or non-preview mode.
* `AbstractJdbcInputPlugin`, the caller of `newSelectCursor`, would determine whether to limit or not. `Exec.isPreview()` would be checked in `AbstractJdbcInputPlugin`, not in `newSelectCursor`.

joe-td: Thanks @dmikurube Got your point. I updated it.

embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/AbstractJdbcInputPlugin.java
```diff
@@ -495,13 +499,22 @@ public TaskReport run(TaskSource taskSource,
 
         LastRecordStore lastRecordStore = null;
 
+        OptionalInt limit = Exec.isPreview()
+            ? task.getPreviewSampleRows()
+            : OptionalInt.empty();
+
+        if (Exec.isPreview() && !task.getPreviewSampleRows().isPresent()) {
+            logger.warn("The preview can be slow due to fetch all records from database");
+        }
```
dmikurube, line 508: Nice warning 👍

embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/PostgreSQLInputConnection.java
```diff
@@ -27,15 +28,23 @@ public PostgreSQLInputConnection(Connection connection, String schemaName, Optio
...
 
         logger.info("SQL: " + query);
         PreparedStatement stmt = connection.prepareStatement(query);
+
+        String fetchSql = "FETCH FORWARD "+fetchRows+" FROM cur";
+
+        if (limit.isPresent()) {
+            stmt.setMaxRows(limit.getAsInt());
+            fetchSql = "FETCH FORWARD "+limit.getAsInt()+" FROM cur";
+        }
```
dmikurube, line 46: How about making it like below so that `fetchSql` is determined just once?

```suggestion
        final String fetchSql;
        if (limit.isPresent()) {
            stmt.setMaxRows(limit.getAsInt());
            fetchSql = "FETCH FORWARD "+limit.getAsInt()+" FROM cur";
        }
        else {
            fetchSql = "FETCH FORWARD "+fetchRows+" FROM cur";
        }
```

dmikurube (approved):
Thanks for working on it! It basically looks good to me, so approved, while leaving a minor comment.

But I'm not very expertised on each RDBMS. Let's ask for some more comments from @hito4t.

embulk-input-postgresql/src/main/java/org/embulk/input/postgresql/PostgreSQLInputConnection.java
joe-td: Thanks @dmikurube I updated it.

hito4t (commented):
It is difficult to evaluate performance by source codes.
I think we should run the program and measure time.

As @dmikurube pointed out, `preview` will not read all records, but at most a few hundred records.
And default `fetch_rows` of embulk-input-jdbc is 10000.
Because it is already small enough for the program and database, `setMaxRows` might have little improvement.

I inserted 10,000,000 records into SQL Server and tried some cases.

embulk run -> 57 seconds
embulk preview -> 10 seconds (340 records shown)
(embulk in my test environment is a bit old (0.9.23) and jruby startup took several seconds.)
read10,000,000 records by JDBC `Statement` -> 22 seconds
read10,000 records by JDBC `Statement` -> 3 seconds

If `SELECT` statement itself will take a lot of time (for example, sort all records by `ORDER BY` clause), I can not come up with a way to improve `preview` performance.

joe-td:
Thanks @hito4t for your comments. You're right, the performance will depend on the query time. So, I think will close this PR.

PR #223, opened 2021-08-19
Remove embulk-input-db2 and embulk-input-oracle

dmikurube:
As we discussed in https://github.com/embulk/embulk-input-jdbc/pull/206#issuecomment-846845438, we're going to remove `embulk-input-db2` and `embulk-input-oracle` from this repository.

PR #258, opened 2024-05-21
jdbc: Explicit rollback before closing connections

ackintosh:
~~Removed the hardcoded autoCommit setting so that users can set it via YAML configuration like this:~~

Added an explicit rollback before closing connections to avoid transaction errors as mentioned [here](https://github.com/embulk/embulk-input-jdbc/pull/258#issuecomment-2123810843).

### DIFF
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
@@ -48,7 +48,6 @@ public class JdbcInputConnection
         if (schemaName != null) {
             setSearchPath(schemaName);
         }
-        connection.setAutoCommit(false);
     }
 
     protected void setSearchPath(String schema) throws SQLException
--- a/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
+++ b/embulk-input-jdbc/src/main/java/org/embulk/input/jdbc/JdbcInputConnection.java
@@ -48,6 +48,7 @@ public class JdbcInputConnection
         if (schemaName != null) {
             setSearchPath(schemaName);
         }
+        connection.setAutoCommit(false);
     }
 
     protected void setSearchPath(String schema) throws SQLException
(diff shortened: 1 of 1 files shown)


dmikurube (commented):
I understand that the behavior hardcoded may not be very good, but just removing this means that a user would lose the compatibility just by upgrading the plugin.

* Could you elaborate what was your real problem with this hardcode, and why removing this would resolve the issue?
* Could you propose a more soft-landing migration path for better compatibility?

dmikurube:
Thanks for your response. Got it. In fact, `embulk-input-db2` was maintained here in the past, but it was removed from this at #223 because we were unable to keep maintaining it (especially, testing it) at that point.

You may want to check the removed code to see how it was working.

Cc: @hito4t

hiroyuki-sato:
JFYI: The transferred repository is https://github.com/hito4t/embulk-input-jdbc-external

Maybe you can try embulk-input-db2 with embulk 0.9.23
And additional info. (Japanese) https://x.com/hito4_t/status/1562697661048049664

ackintosh:
@dmikurube Thanks for the pointer, it helped me a lot!

 I've updated this PR and checked that it works fine locally. I would appreciate it if you could have a look at it.

ackintosh:
Also updated the title of this PR.

hiroyuki-sato:
In my opinion, [d005393](https://github.com/embulk/embulk-input-jdbc/pull/258/commits/d0053938e7b5f67209c96a152630c1e2e2e397a3) is another hardcode. 
Someone probably wants to use explicit commit instead of rollback. It is better to keep current behavior by default, and other behavior implements as optional.

rough ideas.

```yaml
in:
  type: jdbc
  commit_type: none # auto, none, commit, rollback (default none)
```

```yaml
in:
  type: jdbc
  auto_comit: false 
  explicit_commit: # none, commit, rollback
```

ackintosh:
I have decided not to use this plugin for some reasons. I am closing this, but it's okay if anyone wants to reopen or merge it if needed.

@dmikurube @hiroyuki-sato Thanks for your prompt response, which was so helpful. 🙏

PR #264, opened 2024-08-19
Publish from GitHub Actions

hiroyuki-sato:
* copy `.github/workflows/publish.yml` from embulk-input-s3
* Use setup-java and checkout v4 instead of v2.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,32 @@
+name: Publish
+on:
+  push:
+    tags:
+      - "v0.*"
+jobs:
+  publish:
+    environment:
+      name: maven-central-and-ruby-gems
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up OpenJDK 8
+      uses: actions/setup-java@v4
+      with:
+        java-version: 8
+        distribution: "temurin"
+    - name: Publish
+      run: |
+        mkdir -p $HOME/.gem
+        touch $HOME/.gem/credentials
+        chmod 0600 $HOME/.gem/credentials
+        printf -- "---\n:rubygems_api_key: ${RUBYGEMS_API_KEY}\n" > $HOME/.gem/credentials
+        ./gradlew --stacktrace publishMavenPublicationToMavenCentralRepository gemPush
+      env:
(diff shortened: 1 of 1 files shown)


dmikurube:
Please wait a bit. We'll have to set up secrets for releases, and also to change the versioning/releasing strategy (i.e. set `*-SNAPSHOT` version on `master`, and remove it on a detached branch to release).

.github/workflows/publish.yml
```diff
@@ -0,0 +1,32 @@
...
+        distribution: "temurin"
+    - name: Publish
+      run: |
+        mkdir -p $HOME/.gem
+        touch $HOME/.gem/credentials
+        chmod 0600 $HOME/.gem/credentials
+        printf -- "---\n:rubygems_api_key: ${RUBYGEMS_API_KEY}\n" > $HOME/.gem/credentials
+        ./gradlew --stacktrace publishMavenPublicationToMavenCentralRepository gemPush
+      env:
+        ORG_GRADLE_PROJECT_ossrhUsername: ${{ secrets.OSSRH_USERNAME }}
```
dmikurube, line 28: We recently use `vars` instead of `secrets` for username. (See some other repos.)

dmikurube (commented):
Ah, sorry, forgot about it. We may want to release a new version, but as I commented above :

1. Secrets/variables are not set (so it wouldn't work)
2. We'll want to change the versioning/releasing strategy

Can you try (2) when you have time?  Then I'll work on (1).

.github/workflows/publish.yml
hiroyuki-sato: Thanks, I copied these parts.

https://github.com/embulk/embulk-decoder-bzip2/blob/3783be42c7cedc42c3c73c677f3e50451a4d9b8b/.github/workflows/publish.yml#L24-L27

hiroyuki-sato:
The CI of changing [5b87aa1](https://github.com/embulk/embulk-input-jdbc/pull/264/commits/5b87aa10222e19682a6ef7c28fbfb0cea910839e) failed due to a problem with SQL Server. I believe the CI error doesn't relate to this change. (I'll try to fix the CI problem on another PR.)
https://github.com/embulk/embulk-input-jdbc/actions/runs/12665494256

dmikurube:
Yeah, it does not look related, but we may need to dive into it...

hiroyuki-sato:
@dmikurube I fixed the SQL server issue on #271, so I changed the base commit to the latest.

Issue #267, opened 2024-09-12
Remove the jTDS driver.


hiroyuki-sato:
jTDS has been used as Free Microsoft SQL Server Driver. But It doesn't maintain over 10 years. 
Will drop bundle driver in a future.

The latest release was 	2013-06-08
https://sourceforge.net/projects/jtds/files/jtds/

PR #268, opened 2024-09-12
Add the jTDS built-in driver deprecated message

hiroyuki-sato:
Relates to  #267. Remove the jTDS built-in driver in an upcoming release.

### DIFF
--- a/embulk-input-sqlserver/src/main/java/org/embulk/input/sqlserver/SQLServerInputPlugin.java
+++ b/embulk-input-sqlserver/src/main/java/org/embulk/input/sqlserver/SQLServerInputPlugin.java
@@ -225,6 +225,8 @@ public class SQLServerInputPlugin
         }
 
         if (useJtdsDriver) {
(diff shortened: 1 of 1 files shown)


dmikurube:
* It'd be better in general to show such a warning earlier. How about put the log there?
    * https://github.com/embulk/embulk-input-jdbc/blob/6cb64d0f9ac6671df802d2d326ea662e6743c0bb/embulk-input-sqlserver/src/main/java/org/embulk/input/sqlserver/SQLServerInputPlugin.java#L136
* Logging in two lines (two log records) could confuse users when the logging is busy. Let's merge this in a single log line. For example:
    * `"driver_type: jtds" is specified, but note that the jTDS built-in driver will be removed at some point in the future. Please see https://github.com/embulk/embulk-input-jdbc/issues/267`

hiroyuki-sato:
@dmikurube Sorry for late reply. I realized the `driver_path` support needs some source modification. This part also need modify. I'm preparing #269 and #270. Convert to draft for a while.

